PR #8, opened 2011-08-05
2nd try to supply changed rt-message-catalog with ability to process local/ stuff

zito:
Hi,

> From: jibsheet <reply+i-721157-186eb2cb0b546384386836ad8dca9c652767c162@reply.github.com>
> Subject: Re: [rt] 3.8 trunk (#4)

> Unfortunately, this really needs to be done against 4.0-trunk or master to be
> included. Rebasing onto 4.0-trunk causes a number of conflicts because the
> code has already been rewritten to use a +default PO_DIR.

I dove into the GIT yesterday and today and reorganized things a bit. The
original patch against 3.8-trunk is renamed to zito-i18n-1 now. (I was lost
with commits directly in branch 3.8-trunk.) I took yesterdays master branch and
created zito-i18n-2 branch on it. I cherry-pick original commits, but yes there
was a number of rewrites.

I hope these commits are more usable then previous. Thanks for your time to
review.

Best Regards
-- 
Zito


sunnavy:
it's not been updated for years, so I'm closing it for now.

PR #18, opened 2011-11-18
Unify address checking

bobtfish:
This is an attempt to fix an issue I'm having creating tickets by the REST interface - the parsing and validation of email addresses appears overly strict, and there is a FIXME type comment to that effect.

I've changed it to use the generic validation routine in RT::User by splitting a method in half.

The exception we were getting back is:

Cc: glen.d'souza@demonmusicgroup.co.uk is not a valid email address.

which is clearly from the code I've fixed, although I'm unsure that how I've fixed it is correct.

Please take a look.
Cheers
t0m

### DIFF
--- a/lib/RT/User.pm
+++ b/lib/RT/User.pm
@@ -538,6 +538,30 @@ sub LoadOrCreateByEmail {
     return ($self->Id, $message);
 }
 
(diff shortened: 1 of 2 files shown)


sunnavy:
we have an alternative approach to improve this, see 8a34dc3b.

PR #25, opened 2012-05-04
Extract password setting out to its own component

jasonmay:
This component allows a more granular customization by allowing the
customizer to override it for alternative authentication methods.


sunnavy:
it's not completed for quite a while, so I'm closing it for now.

PR #26, opened 2012-05-08
Parse IDs from tags without spaces like [TAG#111]

stbuehler:


sunnavy:
I would agree to allow none spaces too if we haven't already limited to 1+ spaces.
I'm worried that a more loose regex that could possibly break real world somehow.

PR #28, opened 2012-05-17
Queue tag in email from -

bobtfish:
This patch allows you to supply a sprintf string for the 'From' address of queues with tags, allowing you to customise them more than the default


sunnavy:
similar issues, see https://github.com/bestpractical/rt/pull/30#issuecomment-145964863

PR #30, opened 2012-05-22
Subject tag in email to

bobtfish:
Allow per queue tags to be used in email 'To' addresses.


sunnavy:
we already support "the interface needs to correctly support the removal of groups from tickets.", so I assume the first commit is not necessary any more.

the 2nd commit lacks the document of the new option and it seems not quite flexible as subject tag is fixed to be the first %s, so maybe it's good to make a local customization instead of coring it?

PR #42, opened 2012-11-08
Decode MIME headers for better readability in web interface.

petrkle:
Hi,

this change "Subject: =?UTF-8?B?UMWZa....." to readable version.

Cheers
Petr


cbrandtbuffalo:
In current RT 6, this should now be handled for all headers, so the core issue should be resolved.

PR #44, opened 2013-01-19
Fix watcher-transactions' old and newvalue when shredding users

giganteous:
When tickets have transactions that change the Watchers of of it, it
leaves traces in the Transactions table that aren't cleaned properly
when shredding with 'replace_relations'.


sunnavy:
looks good, could you rebase it to current 4.2-trunk? Thanks!

PR #54, opened 2013-03-20
Allow RT::Users->WhoBelongToGroups to optionally return  unprivileged users (4.0)

ringerc:
The WhoBelongToGroups filter should really not apply LimitToPrivileged at
all, but we can't remove it without breaking security assumptions elsewhere
in RT or in 3rd party code. So a new option IncludeUnprivileged is added, a
boolean that defaults to false. It can be set to true to disable the
LimitToPrivileged filter.

The documentation of the method is amended to explain that it filters out
unprivileged members by default.

See http://issues.bestpractical.com/Ticket/Display.html?id=23089

### DIFF
--- a/lib/RT/Users.pm
+++ b/lib/RT/Users.pm
@@ -542,21 +542,31 @@ sub WhoHaveGroupRight
 }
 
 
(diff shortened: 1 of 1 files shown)


sunnavy:
it's already merged.

PR #75, opened 2013-11-19
'Queue' in basic view to url

flashick:
Replaces url instead of the queue name in basic info


sunnavy:
I made 4.2/show-queue-name-link for this, thanks!

PR #79, opened 2013-12-24
Index 'quoted-printable' attachments

yosnoop:
Maybe we can make this optional. Anyhow, I found this very useful in my environment.

### DIFF
--- a/sbin/rt-fulltext-indexer.in
+++ b/sbin/rt-fulltext-indexer.in
@@ -157,7 +157,7 @@ EOT
     }
 }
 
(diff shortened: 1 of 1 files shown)


sunnavy:
IMO, it's good to make an option to support more content types.

cbrandtbuffalo:
RT currently uses the native database full-text indexing, and I wonder how well they would handle the raw =20 type content. Maybe the content could be converted when it is moved to the AttachmentsIndex table. Oracle would need special handling.

PR #83, opened 2014-01-10
Allow grouping by custom field name in addition to id

waldiTM:
In 4.1 you could use GroupBy with a custom field name. The current code filters this out.


sunnavy:
I'm not sure the purpose of this, could you explain more why you want to group by name instead of id?
do you have custom fields that share the same name but with different ids?
with this change, the UI looks quite confusing because you will have 2 options with the same display for each cf.

cbrandtbuffalo:
We didn't get a response on this one with more details, so closing.

PR #84, opened 2014-01-10
Set proper alias for custom field in query

waldiTM:
DBIx::SearchBase sets a column name for each custom field requested in
the query, but this name is neither stable nor really intuitive.  This
patch just sets the name to "CF_$name", with all not applicable
characters replaced.


sartak:
Rather than \W I'd like to see an explicit character class (since I would wonder whether \w was using different semantics, potentially leaving us vulnerable to SQL injection)

PR #91, opened 2014-03-28
fix signature in messagebox

cloos:
By setting $IncludeSignature to 1 the signature is always included in the
message box on replies, regardless of the MessageBoxIncludeSignature
setting.
This fixes the bug by respecting the MessageBoxIncludeSignature setting.


cloos:
@jibsheet never got a response back. It would be great if you can clarify the intended behavior of $MessageBoxIncludeSignature and $MessageBoxIncludeSignatureOnComment.

jibsheet:
@cloos I no longer work for BPS - and I don't have context on this anymore unfortunately.

PR #130, opened 2014-12-04
don't notify disabled owner

cloos:


alexmv:
Reworded and merged; thanks!

PR #134, opened 2015-02-03
fix some rudder styling mistakes

cloos:


alexmv:
The additional `<ul>`s are strange, and result in very wide left margins:

article-search

I'm a little wierded by the loss of the titlebox titles, but I'm fine with the consistency argument there.

cloos:
The ```<ul>``` s just imitates what  Elements/ListActions does.

This produce a wide left margin for the rudder theme (but this is the default Elements/ListActions styling in rudder), but prevents the need to style this for all other themes, as without the ```<ul>```
 s the text is to close to the box border or, for ballard and web2, overlaps with the box title.
screenshot-without-ul

A maybe better (more consistent) solution would be to feed the text as a list to Elements/ListActions.
But as this element doesn't use the title arg (which is used in other places like Elements/Login:54), this would result in a "Result" title instead of the "Setup needed" title.

PR #135, opened 2015-02-05
Update EditCustomField

pccowboy:
Expose Rows and Cols parameters to callbacks.


alexmv:
Looks eminently reasonable.  Reworded slightly and cherry-picked as fca65159, since it looks like your commit history got a little confused.

PR #137, opened 2015-02-10
pass SquelchMailTo to Ticket->Create

cloos:
1eba3f98e9fbd6795b3b4f108c5daba381fbf00a introduced creating tickets
with squelched recepients, but this was only done in Ticket->Create.

Actually the ticket create page passes his args to CreateTicket, which
furthur send only a subset of args to Ticket->Create.

Passing the SquelchMailTo to Ticket->Create makes it possible for custom
ticket create forms or extensions based on RT::Extension::FormTools to
squelch recepients on create.


alexmv:
Moved the `SquelchMailto` line further down, and merged.  Thanks!

PR #139, opened 2015-05-04
Pass a scalar to File::Which::which rather than a list

jmdh:
File::Which 1.17 introduced a scalar prototype for the which() sub,
which broke the previous use of this in RT::Test. See
<https://github.com/plicease/File-Which/issues/6>.

### DIFF
--- a/lib/RT/Test.pm
+++ b/lib/RT/Test.pm
@@ -1718,9 +1718,9 @@ sub file_content {
 }
 
 sub find_executable {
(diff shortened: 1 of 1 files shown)


jmdh:
File::Which has already been fixed (in 1.18) so it may or ma not be still worth workaround around this in RT.

alexmv:
There's little reason not to.  Thanks, merged!

PR #140, opened 2015-05-05
add a new MassageCloneArgs callback

cloos:
This adds the ability to modify the fields that are transferred from the
existing ticket to the new ticket.


sartak:
Thank you cloos!

PR #141, opened 2015-05-12
From apache error_log Warning: Use of "localtime" without parentheses is ambiguous

derekschrock:


sartak:
Merged to 4.2-trunk. Thank you!

PR #142, opened 2015-05-19
Use parens to denote function call

jblaine:
Clears up this:
[Tue May 19 12:35:58 2015] [warning]: Warning: Use of "localtime" without parentheses is ambiguous at /apps/rt4/sbin/../lib/RT/Interface/Email.pm line 525. (/apps/rt4/sbin/../lib/RT/Interface/Email.pm:525)


sartak:
Fixed as #141. Thank you.

PR #144, opened 2015-05-21
fix titlebox title height for rudder

cloos:
The correct padding-bottom of 0.25em is already set in rudder/boxes.css.

This fixes the wrong height of the titlebox title on some pages (e.g.
Ticket Basics) that could only be observed in Firefox.

### DIFF
--- a/share/static/css/rudder/ticket.css
+++ b/share/static/css/rudder/ticket.css
@@ -36,7 +36,6 @@
 .ticket-info-attachments .titlebox-title  .left {
     padding-left: 2.25em;
     margin-left: 0;
-    padding-bottom: 4px;
     margin-bottom: 8px;
     -webkit-border-top-left-radius: 0.3em;
     -webkit-border-top-right-radius: 0.3em;


sunnavy:
with the padding-bottom: 4px, the relative position of titleboxes and the lines below are consistent, but if without it, it's not any more, quite weird. I tested on firefox 41.0.1.
So I think the 4px there was set probably on purpose, @obra, do you still remember this?

obra:
It was trial and error with a range of browsers. 


On Wed, Oct 07, 2015 at 10:29:47AM -0700, sunnavy wrote:
>    with the padding-bottom: 4px, the relative position of titleboxes and the
>    lines below are consistent, but if without it, it's not any more, quite
>    weird. I tested on firefox 41.0.1.
>    So I think the 4px there was set probably on purpose, [1]@obra, do you
>    still remember this?
> 
>    —
>    Reply to this email directly or [2]view it on GitHub.
> 
>    Reverse link: [3]unknown
> 
> References
> 
>    Visible links
>    1. https://github.com/obra
>    2. https://github.com/bestpractical/rt/pull/144#issuecomment-146270171
>    3. https://github.com/bestpractical/rt/pull/144#issuecomment-146270171

--

PR #145, opened 2015-05-26
Try not to decrypt application/pgp-keys attachments

morallo:
Public keys attached with .asc extension (default many mail clients, for example Thunderbird+Enigmail) were wrongly detected as attachments with inline encryption.
This checks the Content-Type: header for application/pgp-keys and omits the decryption.


morallo:
Thanks for the comment. I hadn't seen that branch. Closing the PR then.
Any idea when this will be merged into trunk and released?

alexmv:
Unclear -- depends when a BPS engineer has time to review and merge it.  If it works for you, that'd be helpful to know, I'm sure.

PR #147, opened 2015-05-28
Code warning fix

mauroasilva:
Warning: Use of "localtime" without parentheses is ambiguous at /opt/rt4/sbin/../lib/RT/Interface/Email.pm line 526.


sartak:
Fixed as #141. Thank you.

PR #148, opened 2015-05-29
consistent "now" value on ticket create

cloos:
Without this, creating a ticket with a open status, Started is set to
now in RT::Ticket->Create and Created is set to now in
RT::Record->Create, where the later method call can be a second later
(actually on the next second) than the former one. This results in
Started is one second before Created.

Avoid this by setting Created on RT::Ticket->Create.

Also use a consistent now value for other date values, if they are set
to now.

I'm not sure if we should also include a upgrade step to update Started and Resolved, if they are one second before Created.

### DIFF
--- a/lib/RT/Ticket.pm
+++ b/lib/RT/Ticket.pm
@@ -314,6 +314,10 @@ sub Create {
         unless defined $args{'Priority'};
 
     # Dates
+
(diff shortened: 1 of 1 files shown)


sartak:
Nice catch! Cherry-picked onto 4.2-trunk as d8b700cba4293b3079f11bf786f9e920f9146763. Thanks!

PR #149, opened 2015-06-01
Avoid warnings if users don't have sufficient rights on reminders

cloos:
Users must not have the same rights on tickets and their reminders
(e.g. if you use role based rights for ShowTicket and add a user to the
ticket role but not to the reminder role).

For Transaction->%_BriefDescriptions return a description without the
subject if the user don't have sufficient rights. This fixes warnings
like 'Use of uninitialized value in join or string at
/opt/rt4/sbin/../lib/RT/Transaction.pm line 827'.

For Links grab the Ticket->Type without ACL checks. We check here only
if it's a reminder to avoid display them as links. This fixes warnings
like 'Use of uninitialized value in string eq at
/opt/rtgit/share/html/Elements/ShowLinksOfType line 93'.

### DIFF
--- a/lib/RT/Transaction.pm
+++ b/lib/RT/Transaction.pm
@@ -1257,34 +1257,46 @@ sub _FormatUser {
         my $self = shift;
         my $ticket = RT::Ticket->new($self->CurrentUser);
         $ticket->Load($self->NewValue);
-        my $subject = [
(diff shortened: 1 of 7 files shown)


sunnavy:
merged into 4.2-trunk with little tweaked message, thanks!

PR #39, opened 2012-09-25
new config option @SkipFieldsOnTicketClone

cloos:

PR #154, opened 2015-08-13
configureable cloneticket. You can set what to copy and what not

atus42:
Not everyone needs everything to copied from original ticket to clone ticket when you create a new linked ticket. In our case requestor and timeworked fields cause problem, but I made all the fields configurable. The 'default' is to copy everything as it is now. I created a new sample configuration file, and committed it, but it is not required. If a setting is not defined, it will be copied.


cloos:
@atus42 with the next release, there will be a 'MassageCloneArgs' callback, where you can adjust which fields should be cloned (see aaa07b295e036a3a2f411a07ff8688fe11c7dee2).
This is much more flexible as an config option, you can for example chose which fields to clone by queue (see also #39).

atus42:
Closing because better solution MassageCloneArgs.
Thanks for sharing the info.

PR #155, opened 2015-08-13
form submit on ctrl+enter at creating and updating ticket pages

atus42:
A lot of site has this feature: submitting a form on ctrl+enter. I think it would be a nice feature of RT.

### DIFF
--- a/share/html/Ticket/Create.html
+++ b/share/html/Ticket/Create.html
@@ -45,6 +45,16 @@
 %# those contributions and any derivatives thereof.
 %#
 %# END BPS TAGGED BLOCK }}}
+<script type="text/javascript">
+window.onload = function () {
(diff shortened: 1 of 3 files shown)


cloos:
@atus42 I think new JavaScript shouldn't be in-lined (share/static/js/forms.js would be a better place for your JavaScript).
This makes it possible, once all in-line JavaScript is gone, to use HTTP Content Security Policy.

atus42:
Thanks for the suggestion. I moved the js part to share/static/js/forms.js

cbrandtbuffalo:
This would be a nice shortcut to add. In RT 6, we would want to incorporate it into the other keyboard shortcuts and include it in the help modal so users could find it. We also need to test with htmx since it can interfere with submit events.

PR #156, opened 2015-08-14
configurable attachment ordering method on web ui

atus42:
From time to time we need to attach files to tickets in specified order. Currently RT forcedly orders attachments by name, regardless of the uploading order. To work this around one must rename their files an prepend a suffix to have a correct order (for that ticket). This is a tideious (and endless) work.

Width this patch if a configuration value "SortAttachmentsByUploadingOrder" is exist and the value is not zero, the uploading order of the attachments is preserved.

### DIFF
--- a/lib/RT/Interface/Web.pm
+++ b/lib/RT/Interface/Web.pm
@@ -2151,8 +2151,8 @@ sub CreateTicket {
 
     my @attachments;
     if ( my $tmp = $session{'Attachments'}{ $ARGS{'Token'} || '' } ) {
-        push @attachments, grep $_, map $tmp->{$_}, sort keys %$tmp;
-
+        push @attachments, $tmp->{$_}{ME}
+            foreach sort keys %$tmp;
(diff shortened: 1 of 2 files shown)


sunnavy:
I'm not sure if the feature is useful enough to be included in core.
besides, it fails t/web/csrf.t and doesn't contain document of the new config value SortAttachmentsByUploadingOrder

FYI, in 4.4, we will support multiple files upload via ajax and also include attachments from existing ones, which makes the order a bit more confusing.

atus42:
We use RT about 10-15 years and we usually miss this feature/function. We have pictures, named by content and it may not follow the order they should appear on screen and letter. This is especially true to installation or configuration instructions.

I'll check t/web/csrf.t

I did't know how to properly document the new config value other than putting a description about it to the commit message. Should I put some kind of special comment to the code to 'notify' some script to the new setting?

Unfortunatelly I didn't have time, to follow 4.4 development, so I didn't know about this new feature. I think my modification is pretty simple to follow in case there is no full rewrite in the code at this part.

PR #158, opened 2015-09-01
Skip tests which relate to old behaviour in t/mail/html-outgoing.t

jmdh:
Older versions of HTML-FormatText-WithLinks-AndTables were buggy
and this test assumes this, so skip if a non-buggy version is found.

Bug: https://issues.bestpractical.com/Ticket/Display.html?id=31191


sunnavy:
Merged with tiny tweaked messages.
Thanks!

PR #160, opened 2015-09-09
record transactions for queue changes

cloos:
For recording a Sign, SignAuto and Encrypt transaction, instead of
creating 3 new transaction types, I modified the Enabled and Disabled
types to optionally take a Field arg and record an Enabled/Disabled
transaction with this field instead of the object type.

I also had to anchor the date/time transaction types by the full name to
not match the 'Due' in the queue's DefaultDueIn field.


sunnavy:
merged to 4.2-trunk, thanks!

PR #161, opened 2015-10-15
don't index EmailRecord transaction attachments

cloos:
EmailRecord and CommentEmailRecord transaction attachments contains
redundant content as this attachments consist of the transaction
content (Create, Correspond or Comment) and the template text.

For example with the default RT configuration with queue AdminCcs, a
ticket create results in a Create transaction and two EmailRecord
transactions (one for the Requestor autoreply and one for the queue
AdminCcs).
So the valuable information in the create transaction attachment is
indexed three times.

Tests shows that not indexing the EmailRecord and CommentEmailRecord
transaction attachments drops first index time by 30%, indexed rows by
35%, index data file size by 27% and index index file size by 38%.

Test details (RT 4.2.12, MySQL 5.5.44, 1,534,314 plain/html
attachments):

before:
time /opt/rt4/sbin/rt-setup-fulltext-index --index-type mysql --table AttachmentsIndex
34m58.295s

mysql -BNe 'SELECT COUNT(*) FROM rt4.AttachmentsIndex'
1534314

du -h /var/lib/mysql/rt4/AttachmentsIndex.MY*
1.5G    /var/lib/mysql/rt4/AttachmentsIndex.MYD
782M    /var/lib/mysql/rt4/AttachmentsIndex.MYI

after:
time /opt/rt4/sbin/rt-setup-fulltext-index --index-type mysql --table AttachmentsIndex
24m4.712s

mysql -BNe 'SELECT COUNT(*) FROM rt4.AttachmentsIndex'
1000218

du -h /var/lib/mysql/rt4/AttachmentsIndex.MY*
1.1G    /var/lib/mysql/rt4/AttachmentsIndex.MYD
483M    /var/lib/mysql/rt4/AttachmentsIndex.MYI

### DIFF
--- a/lib/RT/Tickets.pm
+++ b/lib/RT/Tickets.pm
@@ -976,12 +976,19 @@ sub _TransContentLimit {
         }
     } else {
         $self->Limit(
+            ALIAS    => $txn_alias,
+            FIELD    => 'Type',
+            OPERATOR => 'NOT IN',
+            VALUE    => ['EmailRecord', 'CommentEmailRecord'],
+        );
+        $self->Limit(
             %rest,
-            ALIAS         => $self->{_sql_trattachalias},
-            FIELD         => $field,
-            OPERATOR      => $op,
-            VALUE         => $value,
-            CASESENSITIVE => 0,
+            ENTRYAGGREGATOR => 'AND',
+            ALIAS           => $self->{_sql_trattachalias},
+            FIELD           => $field,
+            OPERATOR        => $op,
+            VALUE           => $value,
+            CASESENSITIVE   => 0,
         );
     }
     if ( RT->Config->Get('DontSearchFileAttachments') ) {
--- a/sbin/rt-fulltext-indexer.in
+++ b/sbin/rt-fulltext-indexer.in
@@ -167,6 +167,18 @@ sub attachment_loop {
     {
         # Indexes all text/plain and text/html attachments
         my $attachments = RT::Attachments->new( RT->SystemUser );
+        my $txn_alias = $attachments->Join(
+            ALIAS1 => 'main',
+            FIELD1 => 'TransactionId',
+            TABLE2 => 'Transactions',
+            FIELD2 => 'id',
+        );
+        $attachments->Limit(
+            ALIAS    => $txn_alias,
+            FIELD    => 'Type',
+            OPERATOR => 'NOT IN',
+            VALUE    => ['EmailRecord', 'CommentEmailRecord'],
+        );
         $attachments->Limit(
             FIELD    => 'ContentType',
             OPERATOR => 'IN',
--- a/sbin/rt-fulltext-indexer.in
+++ b/sbin/rt-fulltext-indexer.in
@@ -217,7 +217,7 @@ sub process_bulk_insert {
             debug("Found attachment #". $a->id );
             my $text = $a->Content // "";
             HTML::Entities::decode_entities($text) if $a->ContentType eq "text/html";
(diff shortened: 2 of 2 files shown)


alexmv:
This is https://issues.bestpractical.com/Ticket/Display.html?id=19237 -- and also, it looks like, https://issues.bestpractical.com/Ticket/Display.html?id=30982 .  If we apply this to the FTS indexing, we should also apply it to https://github.com/bestpractical/rt/blob/stable/lib/RT/Tickets.pm#L978-L985 for consistency.

Note, however, that this will cause the simplifying assumption in https://github.com/bestpractical/rt/commit/8450f0a9f233d6a761ac22dbdf14926abc54d7fa to no longer be true -- so searches in FTS-enabled installs will no longer match on subjects, which is likely an unwanted regression.  https://github.com/bestpractical/rt/commit/8450f0a9f233d6a761ac22dbdf14926abc54d7fa details the right fix.

cloos:
I updated my pull request with @alexmv's suggestions.

cloos:
@sartak It would be great if someone could review this PR.

alexmv:
My only hesitation is that this would require re-indexing; otherwise, transactions recorded before an upgrade to this commit wouldn't have their subject in the index, would exclude the EmailRecord records, and thus would no longer show up in results if they'd previously matched only on subject.

I also wonder if we should be using `NOT IN` to exclude those two types, or instead be using `IN` on  Comment / Correspond.  I don't _think_ it makes a difference, but it's worth pondering briefly.

In any case, it's clearly the right direction to go in, but (as always) backwards compatibility rears its head slightly here.  I don't know if that makes it easier to slip into 4.4.1. rather than 4.2.13.

cloos:
> My only hesitation is that this would require re-indexing; otherwise, transactions recorded before an upgrade to this commit wouldn't have their subject in the index, would exclude the EmailRecord records, and thus would no longer show up in results if they'd previously matched only on subject.

Transactions before this commit are already in the AttachmentsIndex table, including the EmailRecord and thus also the subject. So I don't see any search behaviour change if you are not re-indexing. The only change I see is the AttachmentIndex table wouldn’t grow this much as before. ;-)
So you only need to run the re-indexing, if you want a smaller AttachmentsIndex table.

> I also wonder if we should be using NOT IN to exclude those two types, or instead be using IN on Comment / Correspond. I don't think it makes a difference, but it's worth pondering briefly

I chose NOT IN because it's 2 (EmailRecord, CommentEmailRecord) vs. 3 (Correspond, Comment, Create). Also, if ever a new transaction type will be introduced, or was introduced in a local customisation,  which stores attachments, this type will autocratically be indexed.

alexmv:
Ah!  True that the already-built parts of the index already have the subject.  Hesitation withdrawn!

Can you update the commit message with your logic for NOT IN vs IN?  I have no objection to it, but I want it recorded such that it's recorded for posterity.

Other than that, I'm +1 to merge this; @sartak has final say, however.

sartak:
If you can add the NOT IN vs IN explanation, then you've got my +1 as well. I'll merge it into 4.4-trunk for 4.4.1 since re-indexing is not necessary. (If it were then I think it'd have to wait for 4.6)

Thanks for the hard work @cloos!

And thank you for your detailed and careful reviews, @alexmv!

sartak:
Nevermind, since it was just commit message changes, I went ahead and did them (as well as some other typo/grammar fixes). Merged into 4.4-trunk which will become 4.4.1.

Thanks very much @cloos!

PR #165, opened 2015-11-18
fix pt_BR translation of "Open tickets"

terceiro:
As the English string is ambiguous ("the tickets that are open" X "let's open some tickets now"), the original translator used the wrong meaning.


alexmv:
Can you do this in Launchpad?  That's the authoritative source for po files for RT, from which these files in the repository are generated.  https://translations.launchpad.net/rt/4.2/+pots/rt/pt_BR/1324/+translate looks like the locstring in question.

terceiro:
done

PR #169, opened 2015-12-04
silence upgrade messages

cloos:
For 4.3.1 it don't makes sense to display a message for each custom
field during the upgrade (unless you turn on debug logging).

For 4.3.11 this message should only be relevant if you want to debug the
upgrade, but not for the normal user.


sartak:
Thanks Chris! Cherry-picked into 4.4.0-releng :)

PR #170, opened 2016-01-11
rewrite Windows newlines

netsandbox:
Rewrite Windows newlines (CRLF), as seen in recent Outlook 2010.
As the original implementation from 58dfb2b seem to work with LF, making
the CR optional.


djzort:
This was very helpful. Sad it hasnt been merged

djzort:
This still seems like a no brainer.

PR #171, opened 2016-01-12
fix accented queue name (en)coding if not utf8

atus42:
this little fix corrects queue name (en)coding if it contains accented but not necessary utf8 character(s).

The queue name is 'UKK üzemeltetés'.
The line in the (utf8 encoded) aliases file is
ukkuzemeltetes:         "|/usr/bin/rt-mailgate --ca-file /etc/ssl/certs/mithrandir.ca.crt --queue 'UKK üzemeltetés' --action correspond --url https://rt.mithrandir.hu/"

Without the fix, we got the below error is syslog, and the mail waits in the mail queue with a temporary error.

Jan 12 10:44:23 mail RT: [23963] Could not record email: RT couldn't find the queue: UKK Ã¼zemeltetÃ©s
Jan 12 10:44:23 mail postfix/local[25018]: 0BF9B1060: to=<ukkuzemeltetes@mithrandir.hu>, relay=local, delay=0.35, delays=0.01/0.01/0/0.33, dsn=4
.3.0, status=deferred (temporary failure. Command output: RT server error.  The RT server which handled your email did not behave as expected. I
t said:  temporary failure - RT couldn't find the queue: UKK __zemeltet__s  )

The fix should apply to the upcoming 4.4 too, as the file appears to be the same.

### DIFF
--- a/bin/rt-mailgate.in
+++ b/bin/rt-mailgate.in
@@ -56,6 +56,8 @@ use strict;
 use warnings;
 
 use Getopt::Long;
+use utf8;
+no utf8;
 
 my $opts = { };
 GetOptions( $opts,   "queue=s", "action=s", "url=s",
@@ -63,6 +65,7 @@ GetOptions( $opts,   "queue=s", "action=s", "url=s",
             "timeout=i", "verify-ssl!", "ca-file=s",
           );
 
+if ( !utf8::is_utf8( $opts->{"queue"} ) ) { utf8::decode( $opts->{"queue"} ); }
 my $gateway = RT::Client::MailGateway->new();
 
 $gateway->run($opts);


cloos:
@atus42 you can also use the queue id instead of the queue name to workaround your problem.
This also have the benefit that you don't have to change the alias file if you change a queue name.

atus42:
Thanks for the reply. Queue ID-s are not as 'readable' as queue names, but it is good to know that we can use id-s too.

atus42:
We tested the utf8 / non utf8 part, and indeed it was not necessary. None of them. (we saw this code elsewhere and just used it)

We tried the Encode::decode("utf-8", ...) method, but it didn't even code the queue name to utf8.

We did more tests to see what happens in case of real utf8 characters, like ő, and found out that our initial 'solution' failed this test. On the other hand it showed that lwp::useragend didn't handle utf8 characters in queue name form-data and searched another solution.

Our tests shows that encoding/decoding queue name with HTML::Entities solves all problems (we know about).

I committed the new code. We tested it with letters áé, ő and 'english only' in queue names and éáűőúöüóíÉÁŰŐÚÖÜÓÍ letters in the mail body. The ticket(s) got created in every case and the mail body contained all the letters with proper coding.

PR #176, opened 2016-01-27
Fix case if Issuer  DN does not contain email field

mkosmach:


mkosmach:
I have another idea - we can always use full DN for certificates without email filed in DN.
In this case RT will correctly show  user certificates without email in DN too.

PR #178, opened 2016-02-04
4.2/add txn rest customfield

puck:
These patches allow CustomFields to be set on transactions via the REST API.


puck:
Update to fix a warning on Perl 5.26. Rebased onto 4.2-trunk as of today, and also tested successfully against 4.4-trunk.

djzort:
At this point, id be glad for either/any to be pulled. Im not sure why its not being pulled in?

djzort:
Still just sitting here. Sad.

PR #182, opened 2016-03-16
Added merged ticket IDs to ticket details in REST API.

muganor:
We are writing a synchronization job between RT and other systems. We use queries to find tickets updated/created since last sync time.
Then, when querying tickets using their IDs, merged tickets will instead return the "new" ticket information.  We must be able to detect merged tickets and we cannot afford to query information about all existing open tickets at all time to discover merged tickets and merge them in the other systems.  
We could query the merged ticket history but it seems a bit too expansive and exhaustive only to detect merge operations.
Exposing the list of ticket IDs merged into a given ticket gives the same information more efficiently.

### DIFF
--- a/share/html/REST/1.0/Forms/ticket/default
+++ b/share/html/REST/1.0/Forms/ticket/default
@@ -236,6 +236,8 @@ if (!keys(%data)) {
     my ($time, $key, $val, @data);
 
     push @data, [ id    => "ticket/".$ticket->Id   ];
+    push @data, [ Merged => join(", ", $ticket->Merged) ]
+        if $ticket->Merged;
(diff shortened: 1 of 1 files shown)


cbrandtbuffalo:
We're reviewing our PR backlog and we're aware that some are very old. Sorry the the long delay.

For this one, we are very hesitant to make changes to the REST 1 interface because it has been in use for a very long time. Adding something new has the potential to break some old integrations. We're open to changes to REST 2 code.

For this case, I wonder if you could check the the ID you queried with and the id that was returned? If they are different, then you know the ID you queried was merged into the ID you got back.

PR #184, opened 2016-04-09
force .load to POST instead of GET

trwww:
AJAXifying the preview scrips with .load() via .serialize() does a GET
request with the Display.html form, and the query string can end up too
long for the server to accept it.

This commit adds a helper to serialize the form as an object instead of
query string, which tells .load to POST instead of GET. It looks like
theres no built in jQuery interface to serialize a form as an object so
the helper is needed.

Fixes issues#31874

### DIFF
--- a/share/html/Ticket/Update.html
+++ b/share/html/Ticket/Update.html
@@ -222,7 +222,7 @@ jQuery( function() {
            }
        );
        jQuery('#previewscrips div.titlebox-content').load( '<% RT->Config->Get('WebPath')%>/Helpers/PreviewScrips',
-           jQuery('form[name=TicketUpdate]').serialize(),
+           jQuery('form[name=TicketUpdate]').serializeObject(),
            function() {
                var txn_send_field = jQuery("#previewscrips input[name=TxnSendMailTo]");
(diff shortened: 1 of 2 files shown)


JvdW:
I implemented this patch and it works OK but has a unpleasant side effect.
How to reproduce:
Reply to a ticket and make sure you have the right to see the scrip preview pane
Notice that all that need to recieve email are having a checkmark!
Click into the message body reply field, type some text, not really needed
Click outside the text field on a place that is NOT a other form element such as the Owner dropdown.
Notice that the checkmarks disappear :-(

alexmv:
I commented on the issues ticket, but not here.  [4.4/preview-scrips-post](https://github.com/bestpractical/rt/compare/4.4-trunk...4.4/preview-scrips-post) fixes both of these issues.

trwww:
Closing pull request as @alexmv's branch is much cleaner.

sartak:
Hi Todd, JvdW, Alex. Just to close the loop on this, I merged Alex's 4.4/preview-scrips-post branch to 4.4-trunk as 63944d748204abfb1e4b69663481cd098d067a00. Thanks everybody. :)

PR #185, opened 2016-04-11
Added ability to perform searches including deleted tickets.

muganor:
We are writing a synchronization job between RT and other systems. We use queries to find tickets updated/created since last sync time.
Unfortunately, there is no way using the search REST API to query for recently deleted tickets. Polling status for each previously synced ticket is out of the question since it is inefficient and volume would produce too much server load.

This adds a new query parameter "include_deleted" to enable search support for deleted tickets only.

I didn't know the search use-cases for other entity types (ie users, queues) and if you provide pointers I can also make my patch support them, although it seems there are existing "LimitToDisabled/Enabled" calls for those entity types.

### DIFF
--- a/share/html/REST/1.0/search/dhandler
+++ b/share/html/REST/1.0/search/dhandler
@@ -52,6 +52,7 @@ $query
 $format => undef
 $orderby => undef
 $fields => undef
+$include_deleted => undef
 </%ARGS>
 <%INIT>
 my $type = $m->dhandler_arg;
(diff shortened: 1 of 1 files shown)


cbrandtbuffalo:
As noted on [https://github.com/bestpractical/rt/pull/182](url), we're not actively making changes to the REST 1 interface. With the REST 2 interface available in newer RT, you can query transactions. Using that, you could send a request for something like "Created > '1 hour ago' AND NewValue = 'deleted'" to find all deleted tickets.

PR #186, opened 2016-04-12
Inverting order of recommended perl webservers

rachelkelly:
Inverting order of recommended perl webservers in order to guide people to FastCGI instead of mod_perl first.


sartak:
Cherry-picked into 4.2-trunk as dc0992c7c5df2a570efaf898bea766969e7b5a40. Thanks!

PR #187, opened 2016-05-06
rt-ldapimport: add POD

jmdh:
Convert usage to POD, and add a reference to the full documentation at
RT::LDAPImport.


jmdh:
This seems to have been added independently in 9dcd1a04f72230e802d13e5ee9279dbac08af65e, closing.

PR #189, opened 2016-05-11
ignore From_ line in header

netsandbox:
MDA (Postfix, procmail) add a From_ line to the header like this:

From cloos@netcologne.de  Tue May 10 14:55:10 2016

If we don't ignore this line ContentAsMIME split this line by the first
':' and then reassembles the line to a header like this:

From cloos@netcologne.de  Tue May 10 14: 55:10 2016

This isn't RFC 5322 conform, because header names aren't allowed to
include spaces.

MIME::Parser up to 5.428 already ignored the From_ line, but this was
removed in 5.500.

### DIFF
--- a/lib/RT/Attachment.pm
+++ b/lib/RT/Attachment.pm
@@ -769,6 +769,7 @@ sub _SplitHeaders {
     # continuations and rely on this bogus split pattern, however, so it is
     # left as-is for now.
     for (split(/\n(?=\w|\z)/,$headers)) {
+        next if /^>?From /;
         push @headers, $_;
 
     }


alexmv:
rt.perl and rt.cpan [suffer from this](https://rt.perl.org/Public/Bug/Display.html?id=128134&ShowHeaders=1), and IIRC there's a ticket about it somewhere, possibly tickets.bps.  I think we failed to replicate it, not realizing the MIME::Parser version dependency.

I don't see anything in the [changelog](https://metacpan.org/changes/distribution/MIME-tools) about it, but sure enough, it was [silently removed in 5.500](https://metacpan.org/diff/file?target=DONEILL%2FMIME-tools-5.500%2F&source=DONEILL%2FMIME-tools-5.428%2F#lib/MIME/Parser.pm), having been there since version 5 was released in 2000.

This is a looser form of what MIME::Parser had, since it allows the "From " header to appear any where in the list of headers.  This is _probably_ fine, but a better fix might be to mangle `@headers` after the split using  `shift @headers if @headers and $headers[0] =~ /^>?From /`.

However, I don't buy that plain postfix adds a `From_` header: see [the manpage for the pipe transport](http://www.postfix.org/pipe.8.html), under `flags`, which notes that "by default, a message is copied unchanged`, and cites `F` as a flag to add the `From_` header.

Procmail's docs are much muddier.  But I fear that it's not just prepending the line, but rather sending over the pipe to `rt-mailgate` as [mbox](https://en.wikipedia.org/wiki/Mbox), with all of the quoting of `From ` lines in the body that that entails.  It looks like that only kicks in in the absence of a `Conent-Length` header, but I'm not sure.  If you could do some testing on this, to see if we need to care about Procmail having munged the body, that'd be appreciated.

I'd cite the source code for Procmail to help determine its behavior, but I just looked at it and my eyes started bleeding.  Among other things, it rolls its own `strstr()` from an old version of glibc (including the "I deliberately chose not to comment it.  You should have at least as much fun trying to understand it, as I had to write it. :-)" comment) and maintains about that level of comprehensibility throughout.  Determining its behavior experimentally is certainly the easier way to go.

cloos:
>This is a looser form of what MIME::Parser had, since it allows the "From " header to appear any where in the list of headers. This is probably fine, but a better fix might be to mangle @headers after the split using shift @headers if @headers and $headers[0] =~ /^>?From /.

The From_ line isn't always the first line in @headers (don't know why the headers are not in the same order as the MDA sends it to RT).

>However, I don't buy that plain postfix adds a From_ header: see the manpage for the pipe transport, under flags, which notes that "by default, a message is copied unchanged, and citesFas a flag to add theFrom_` header.

We use a default Debian Postfix install and see the From_ line. We can't figure out which Postfix service is used if you have set-up the aliases as described in README 10). Maybe you know this?

PR #190, opened 2016-07-06
fix fetching user for MailCommand mbox

cloos:
With fastcgi deploys, %ENV doesn't have a USER key.
Use getpwuid here to fetch the user.

This fixes warnings:
Use of uninitialized value $ENV{"USER"} in concatenation (.) or string


sartak:
Hey Chris,

Committed as 03290531c157f9a26a73bc2c915e672c12e2ed80 onto 4.2-trunk. I amended your commit a little to continue using `$ENV{USER}`, adding `getpwuid($<)` as a fallback.

Cheers!
Shawn

PR #191, opened 2016-07-19
Skip the password change form with external auth.

smithj4:
This adds the same *NO-PASSWORD* check that the internal Prefs/AboutMe.html page has. For RT installs that are using $WebRemoteUserAuth or $ExternalAuthPriority, there is no internal RT password and the presence of the password change form is confusing to SelfService users.


sartak:
Hi Jason,

I've cherry-picked this into 4.4-trunk as e78766c9d29e7010a95b64b6297baf9287e8309f. It'll be in RT 4.4.2 and beyond.

Thanks!
Shawn

PR #193, opened 2016-08-26
add a new $OverrideMailPrecedence config option

cloos:
While you can set a specific Precedene header for a queue by a queue
template, as described in $DefaultMailPrecedence, you can't supress a
Precedence header this way.
Testing shows, that sending an mail with an empty Precedence header, the
responder don't send automatic responses, like vacation messages.


sartak:
Merged into 4.2-trunk as c9774928c6da3757b266e305a2794f0b4122ac47. Thanks Chris!

PR #194, opened 2016-08-27
Fixes relating to gpg 2.1 (in)compatibility

jmdh:
RT doesn't work with gpg 2.1 at the moment (mostly because GnuPG::Interface doesn't. This patch series fixed this indirectly in Debian to not hardcode the path to gpg by default, so that the default configured in GnuPG::Interface is used instead.

### DIFF
--- a/etc/RT_Config.pm.in
+++ b/etc/RT_Config.pm.in
@@ -2525,7 +2525,8 @@ be found by running the command `perldoc L<RT::Crypt::GnuPG>` (or
 Set C<Enable> to 0 or 1 to disable or enable GnuPG interfaces
 for encrypting and signing outgoing messages.
 
-Set C<GnuPG> to the name or path of the gpg binary to use.
+Set C<GnuPG> to the name or path of the gpg binary to use to override
+the GnuPG::Interface default.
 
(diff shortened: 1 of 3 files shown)


jmdh:
I've just remove the third commit from this PR, as it was clearly specific to Debian; sorry about that.

alexmv:
Removing the `->Probe` call doesn't just affect GnuPG -- that codepath is also used for SMIME.

jmdh:
This is true; run-time probes are less important for Debian as the package dependencies should enforce the presence of the necessary tools but I can see how you wouldn't want to remove it completely.

Perhaps it should be a configuration option - maybe even an autoconf one.
Of course the best course of action would be GnuPG::Interface [being fixed to work with gpg 2.1](https://rt.cpan.org/Ticket/Display.html?id=102651) - in that case, the immediate need for this patchset would be gone (although personally I think it's still worthwhile in case someone wants to configure a particular gpg for some other reason).

jmdh:
This is being reverted in Debian - it didn't turn out to be useful because we need to have gpg in the test suite and in GnuPG::Interface configured explicitly. See

PR #195, opened 2016-09-06
Fix RT::I18N and the test suite to work without cwd in @INC

jmdh:
This change to remove '.' from @INC has been made in Debian recently, and is expected to be made upstream for perl 5.26; see https://rt.perl.org/Ticket/Display.html?id=127810

### DIFF
--- a/lib/RT/I18N.pm
+++ b/lib/RT/I18N.pm
@@ -56,6 +56,7 @@ package RT::I18N;
 
 use strict;
 use warnings;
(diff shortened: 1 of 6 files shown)


sartak:
Thanks Dom and Niko. Cherrypicked into 4.2-trunk as b00a27fdb942e033bc6eaa83cac228298da425c0. It'll be merged up into 4.4-trunk for the 4.4.2 release.

PR #198, opened 2016-10-17
Text attachement from external storage with right encoding

petrkle:
Fix text encoding on attachments from external storage.


petrkle:
Yes, you right.

alexmv:
I've pushed [4.4/external-encodings](/bestpractical/rt/tree/4.4/external-encodings) based on this.  It uses your tests as a base (crediting you in the commit message), but cleans them up, and expands on the origins of the bug and failure modes in the commit message.

PR #201, opened 2016-11-14
Add TicketSQL placeholder for username

jmmcatee:
I have seen a lot confusion around the ability to reference just the User->id in TicketSQL. We are using this parse tweak to compare based on the username as well. Not sure if you all except pull requests like this. let me know if there is a different way I should go about doing this.

### DIFF
--- a/lib/RT/Tickets.pm
+++ b/lib/RT/Tickets.pm
@@ -3116,6 +3116,9 @@ sub _parser {
 
             # replace __CurrentUser__ with id
             $value = $self->CurrentUser->id if $value eq '__CurrentUser__';
+            
+            # replace __CurrentUsername__ with the username
+            $value = $self->CurrentUser->Name if $value eq '__CurrentUsername__';
 
             my $sub = $dispatch{ $class }
                 or die "No dispatch method for class '$class'";


sartak:
Hi @jmmcatee,

We'd be happy to apply your patch but I'm a bit confused about why it's necessary. Could you explain what you are doing that requires a `__CurrentUsername__` variable? As far as I can tell, TicketSQL  handles `__CurrentUser__` being an id rather than the username fairly well.

Thanks,
Shawn

jmmcatee:
Shawn,

It is possible we are an edge case here but we integrate Active Directory for user account management with RT. We have various queues that deal with requests related to users. Some for instance are requesting a list of users be added to a group. To accomplish this we built to external data plugins to auto complete group names and user names. Our people want to be able to see all tickets they have created or are assigned to in our custom fields that utilize the external values from Active Directory. The only way I could figure out how to do this was to get the username of the logged in user and compare that to the custom field value in a saved search. 

Does this help and make sense? Am I missing a way to do this without all the external values and custom field?

Michael

sartak:
Hi Michael,

Thanks very much for your contribution. I've merged it into 4.4-trunk as 575212e054a40df1a71b6e796a35bf2dbaa326d8 with some minor adjustments (I changed the casing from `__CurrentUsername__` to `__CurrentUserName__`, and removed spurious whitespace). It'll be included in the next release, RT 4.4.2.

Best,
Shawn

PR #202, opened 2016-11-27
Adapt SMIME probe to work with openssl 1.1

jmdh:
OpenSSL 1.1 dropped the list-standard-commands command, so also try
'openssl list -commands'.

This has been tested in both Debian unstable (which as of the start of November has OpenSSL 1.1) and stretch (which still has OpenSSL 1.0).


sartak:
Thanks Dom!

PR #203, opened 2016-12-29
Add autocomplete for assets

gibus:
Add support of autocompletion for assets (with RT::Assets-SimpleSearch,
therefore using RT->Config->Get('AssetSearchFields') and
RT->Config->Get('DefaultCatalog')) in field "Add an asset to this
ticket" of the ticket display form.

Also add support of completion both for tickets and for assets in each
input field of the AddLinks form. Completion is done for tickets (as
usual) until the entered term starts with "asset:", where the completion
is switched for assets.

Signed-off-by: gibus <gibus@easter-eggs.com>

(NB: I've previously submitted this patch on https://issues.bestpractical.com/Ticket/Display.html?id=29204 but was advised by @elacour that code review may be easier with a pull request on github)

### DIFF
--- a/share/html/Elements/AddLinks
+++ b/share/html/Elements/AddLinks
@@ -55,8 +55,9 @@ my $id = ($Object and $Object->id)
     ? $Object->id
     : "new";
 
-my $exclude = qq| data-autocomplete="Tickets" data-autocomplete-multiple="1"|;
-$exclude .= qq| data-autocomplete-exclude="$id"| if $Object->id;
+my $exclude = qq| data-autocomplete="TicketsAssets" data-autocomplete-multiple="1"|;
+my @excludes;
+push @excludes, $id if $Object->id;
 </%init>
 % if (ref($Object) eq 'RT::Ticket') {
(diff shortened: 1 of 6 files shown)


mnadai:
Hello gibus

I'm working on something similar and run into your patch.

Have you considered adding the ability to autocomplete based on asset name and/or description?

After all, tickets can be looked up by ticket# or subject, and most likely an asset number won't mean much to whoever is working on a ticket.

Thanks
Maurizio

gibus:
Hi Maurizio,

You can do this through the $AssetSearchFields configuration parameter (see documentation in RTConfig.pm).

For instance, with :
```
Set($AssetSearchFields, {
    Name        => 'LIKE',
    'CF.{Serial number}' => 'STARTWITH',
});
```
the autocompletion is based on Name and the custom field 'Serial number'.

ktm-github:
Hi,
I have patched my RT 4.4.4 instance it it worked once I set the DefaultCatalog but it is not sticky and the only catalog you can use is that one even though the simple search is sticky once you change to a different default catalog, this autocomplete does not. Do you have any ideas?

Regards,
Ken

PR #204, opened 2016-12-29
Add Conditional CustomFields

gibus:
Allows to set up any Custom Field to be conditional. A conditional CF is attached to an object (ticket, queue, user, group, asset, article) if and only if the condition is met for this object. As for categories, a conditions is set up by specifying another Select Custom Field attached to the same object and a value which this Select CF should have for this condition to be met.

### DIFF
--- a/docs/initialdata.pod
+++ b/docs/initialdata.pod
@@ -240,6 +240,18 @@ value.  A clever way to do this easily is with a simple variable you increment
 each time (as above with C<$i>).  You can use the same variable throughout the
 whole file, and don't need one per CF.
 
+=item C<ConditionedBy>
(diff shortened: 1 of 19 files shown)


gibus:
You can forget about this pull request, sorry for this. The functionnality would be better in an extension. So I've just uploaded it, rewritten as an extention, with some tests, documentation, cleaner api and dynamic (javascript to toggle CondutionedBy CFs when Condition CF changes) behaviour: https://metacpan.org/pod/release/GIBUS/RT-Extension-ConditionalCustomFields-0.02/lib/RT/Extension/ConditionalCustomFields.pm

PR #205, opened 2017-01-03
Add POD for rt-ldapimport

jmdh:
Adding POD for consistency with other scripts, and to provide a manpage for Debian.


jmdh:
Ignore, this is a duplicate of #187 which I created the last time I tried packaging RT 4.4. I managed to lose this commit when rebasing so recreated it. Sorry about the noise.

PR #207, opened 2017-02-17
pass along more arguments to SelectDate callbacks

mzagrabe:
This allows callbacks to modify the ShowTime argument for SelectDate
based on things like the Ticket's Queue.


sartak:
Thanks Matt!

PR #211, opened 2017-03-24
Fix logging of users IP address when RT is hosted behind a reverse proxy

wheldom01:
Patch ensures that the clients ip address is logged if the proxy is configured to pass the information using the widely used X_FORWARDED_FOR header.


sunnavy:

-sunnavy

PR #213, opened 2017-03-24
LDAPImport start_tls

wheldom01:
Patch adds support for the use of Net::LDAP start_tls to the LDAPImport.

### DIFF
--- a/lib/RT/LDAPImport.pm
+++ b/lib/RT/LDAPImport.pm
@@ -68,6 +68,8 @@ RT::LDAPImport - Import Users from an LDAP store
 In C<RT_SiteConfig.pm>:
 
     Set($LDAPHost,'my.ldap.host');
+    Set($LDAPTLS, {verify => 'require',
+                   cafile => '/etc/ssl/certs/CACert.pem'});
     Set($LDAPUser,'me');
     Set($LDAPPassword,'mypass');
     Set($LDAPBase, 'ou=People,o=Our Place');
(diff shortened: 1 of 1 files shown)


sunnavy:
Sorry for the late response.

I have 2 main concerns:

* it looks like Net::LDAP::start_tls accepts hash instead of hashref

but in your code:

    $msg = $ldap->start_tls( $RT::LDAPTLS );

I'm wondering if it works.

* $msg returned by start_tls isn't checked

We probably need to check it to see if there are any errors, just like the way we check $ldap->bind.

Thanks!

wheldom01:
Hi,

No problem, I'll take a look as this this evening as it is a while since 
I wrote this patch.

Best Regards

Martin

tobiasbp:
It would be great if this could be completed and merged in. Thanks.

tobiasbp:
Is this change compatible with RT 4.4.4? This is the message I receive when running it:
`Unable to run rt-ldapimport without dependencies.
Rerun configure with the --enable-externalauth option`

wheldom01:
Hi,

I believe that it should still be compatible, but can't confirm that as 
I'm not running 4.4.4 yet.
Could you confirm that the LDAPTLS variable is set correctly?

Set($LDAPTLS, {
     'verify'            =>  'require',
     'cafile'            =>  '/etc/ssl/certs/CACert.pem', } );  # Path CA 
file

I'll try to test it over the weekend.

Best Regards

Martin

wheldom01:
Hi,

You will need the following modules installed to.

Crypt::SSLeay
IO::Socket::SSL
Net::SSLeay

Best Regards

Martin

cbrandtbuffalo:
In 4c288fce, we ended up modifying LDAPImport to allow passing any arguments supported by Net::LDAP, so this is now covered. We also added TLS options to the example so it's clear how to configure a secure connection. This is available in RT 4.4.4 and newer, including RT 5.

Thanks for opening a PR and suggesting the improvement!

PR #214, opened 2017-03-28
Need a new Callback for RT-Extention-AttachmentZip

tbrumm:
Hi Bps,
i'm rewriting the old RTx::AttachmentZip at the moment to get it running under RT 4.4. To make it in a clean way, i would like to request a new Callback for ShowAttachments called AfterAttachmentTitelBox.

Hope it is ok this way.

### DIFF
--- a/share/html/Ticket/Elements/ShowAttachments
+++ b/share/html/Ticket/Elements/ShowAttachments
@@ -51,7 +51,7 @@
         color => "#336699",
         hide_chrome => $HideTitleBox &>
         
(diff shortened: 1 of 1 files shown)


sartak:
Hi Torsten,

I cleaned up your commits and committed to 4.4-trunk as e5333f93a2a544b541cca662abcc06ac23d8785b. Note that I renamed the callback to BeforeList (and added a second one) and passed many more parameters to it.

Thanks very much!
Shawn

PR #215, opened 2017-04-04
4.4/group name input size

mzagrabe:
BPS,

In talking with Jim we discussed some of these changes. The main divergence is extending /Elements/EditCustomField to take a $Classes parameter to extend the class attribute of the rendered HTML for the /Elements/EditCustomField* components.

I couldn't bring myself to sully the component call (in /Admin/Groups/Modify.html) with "Cols => X" when everything else was in CSS.

Let me know what you think.

Also, somehow Shawn's commit (https://github.com/mzagrabe/rt/commit/1846c153c2fe5a294500660f1d4d36735f4c2d5a) got included in this PR. <confused/>

-m

### DIFF
--- a/lib/RT/Interface/Web.pm
+++ b/lib/RT/Interface/Web.pm
@@ -4522,6 +4522,50 @@ sub GetCustomFieldInputNamePrefix {
     RT::Interface::Web::GetCustomFieldInputNamePrefix(@_);
 }
 
+=head2 HTMLifyClassAttribute
+
+Given a parameter list (of scalars or array references), return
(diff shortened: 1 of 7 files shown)


cbrandtbuffalo:
@mzagrabe I think most of the issues touched on here have been addressed in various ways, some with the update to Bootstrap. We don't have the specific classes implementation for custom fields. Do you still have a use case for that in what you have seen of RT 6?

mzagrabe:
@cbrandtbuffalo - I don't recall the reasoning for this PR. Closing! :)

PR #216, opened 2017-04-06
Use all SquelchMailTo on Ticket Create

bconry:
originally from [bestpractical.com #174642]

The _SquelchMailTo method only uses the first value in @_ and silently discards the rest


cbrandtbuffalo:
This also got reported on our [public RT](https://rt.bestpractical.com/Ticket/Display.html?id=31600) and was fixed in b9a3f929. Similar fix, but up the stack a bit in the method itself. The issue is fixed in RT 4.4.2 and newer.

Thanks for the PR and pointing out the issue!

PR #217, opened 2017-04-11
Display Lifecycle Name at Quicksearch/Queue List

tbrumm:
Hi BPS,
not sure if you agree, but i think it is great to display the Lifecycle also at the Queue List.

Torsten


sartak:
We display lifecycle name in RTIR, so I'm on board with this. I'd even go one step further and use:

    <%$lifecycle->Name%>

rather than

    Queue for (<%$lifecycle->Name%>)

since "Queue for" just uses up valuable space

tbrumm:
Cool! Sound much better to me!

sunnavy:

-sunnavy

PR #218, opened 2017-04-13
File is missing during upgrade from older versions

tbrumm:
See Forum: https://forum.bestpractical.com/t/is-there-a-way-to-skip-indexes-during-upgrade/31819/4

### DIFF
--- a/etc/upgrade/3.7.81/schema.mysql
+++ b/etc/upgrade/3.7.81/schema.mysql
@@ -1,0 +1,1 @@
+CREATE INDEX CachedGroupMembers3 on CachedGroupMembers (MemberId, ImmediateParentId);


cbrandtbuffalo:
> See Forum: https://forum.bestpractical.com/t/is-there-a-way-to-skip-indexes-during-upgrade/31819/4

Hi @tbrumm , the method `MakeSureIndexExists` called in that indexes file tries to create the index if it is not found. We use this sometimes when adding new indexes that some systems might already have, maybe because it was discussed publicly and it was suggested as an improvement, so some number of DBs would already have added it. For the reporter on the forum, it looks like it did try to run, but it's not clear why it didn't work. So the fix at that point was something on that system, but adding the index is covered by that indexes file.

PR #221, opened 2017-05-05
callback after table opens for Ticket/Create.html

mzagrabe:
BPS,

I have a need for a callback before the requestors table row on Ticket/Create.html. This patch should do it. I'm open to a different name for the callback and/or a different placement of the callback in the DOM as long as I can get a TR before the Requestors TR.

Let me know what you think.

-m


mzagrabe:
Looks like Brian merged (93b6ebaef) this branch. Closing PR.

PR #223, opened 2017-05-26
add CreatorObj reference to Default callback

mzagrabe:
Use a lexical scope variable for CreatorObj and expose it to default
callback via a reference argument.

This allows admins to modify the Creator that gets displayed for the
transaction.


sartak:
Thanks, merged to 4.4-trunk!

PR #225, opened 2017-06-12
Improved User-Interaction for Attachments

BubblesToTheLimit:
The Dropzone Function expects MiB, not KiB, therefor the previous 10000 would result in an upload limit of 10GiB.
Setting it to 10 actually gives the user a feedback that his attachment size is too big, with the previous code the attachment just gets dropped due to the backend crashing (mysql can only handle a limited package size) and leaves the user puzzled.

The best way would be to get the $MaxAttachmentSize setting from the user config and correct its unit, but my attempts failed. This is the best approach i have so far:
<% use integer; RT->Config->Get('$MaxAttachmentSize') / 1024 %>

### DIFF
--- a/share/html/Elements/JavascriptConfig
+++ b/share/html/Elements/JavascriptConfig
@@ -49,7 +49,7 @@
 my $Config = {};
 $Config->{$_} = RT->Config->Get( $_, $session{CurrentUser} )
   for qw(rtname WebPath MessageBoxRichText MessageBoxRichTextHeight
-         MessageBoxUseSystemContextMenu);
+         MessageBoxUseSystemContextMenu MaxAttachmentSize);
 
 my $CurrentUser = {};
 if ($session{CurrentUser} and $session{CurrentUser}->id) {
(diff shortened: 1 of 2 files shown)


sartak:
Hi @BubblesToTheLimit, thanks for your contribution!

The standard way to expose RT config settings to JavaScript is in the /Elements/JavaScript file here: https://github.com/bestpractical/rt/blob/rt-4.4.2rc1/share/html/Elements/JavascriptConfig#L50-L52

Then to use it, check `RT.Config.Whatever` like so: https://github.com/bestpractical/rt/blob/rt-4.4.2rc1/share/static/js/util.js#L297

Instead of `use integer` it would be better to use the `int(...)` routine, since loading an entire module to change the behavior of one division is awfully overkill :)

Want to give it a shot?

BubblesToTheLimit:
Thanks for the quick response, i adapted your suggestions in these two new commits.
It works like a charm in my installation, now i can change the upload size without clearing the mason cache ;)

sartak:
Hi @BubblesToTheLimit,

Thanks very much for your contribution. I combined it all into one patch and simplified a few things. For example Dropzone can handle fractional values, so it's better not to round (e.g. if limit is 500KiB, we don't want to round that to 0). Merged to 4.4-trunk as 949b295f1bb746110221855ebf87c32b86e91347.

Thanks!
Shawn

PR #226, opened 2017-06-19
Fix "Unescaped left brace in regex" warning in perl 5.26

jmdh:
Originally reported in Debian as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=865047


sartak:
Thanks, merged to 4.4-trunk!

PR #228, opened 2017-07-25
Add Description as tooltip to CF

mkosmach:


cbrandtbuffalo:
Custom fields now have a dedicated field "Entry Hint" that is shown as a tooltip, so this is covered.

PR #229, opened 2017-08-08
Update Ticket.pm, allow Forward From in ARGS

MarkHofstetter:
this minor change allows the From Address for Forwards to be passed in in args

if "From"  is set it is used as the mail From Address in the mail header

if it is not set the Default behavior is not touched

I use this to set the From address to be set as desired by the user in 
https://github.com/MarkHofstetter/RT-Extension-SelectBoxForwardFrom


sunnavy:

-sunnavy

PR #230, opened 2017-08-10
Fix error thrown when adding one-time CC.

gordonmessmer:
This change fixes the first bug discussed here:
http://requesttracker.8502.n7.nabble.com/Onetime-CC-are-giving-an-internal-error-td61678.html

[9625] [Thu Apr  7 12:37:48 2016] [warning]: Use of uninitialized value in split at /opt/rt4/share/html/Helpers/PreviewScrips line 81. (/opt/rt4/share/html/Helpers/PreviewScrips:81)
[9625] [Thu Apr  7 12:37:48 2016] [error]: Operation "eq": no method found,
        left argument has no overloaded magic,
        right argument in overloaded package Email::Address at /opt/rt4/share/html/Helpers/PreviewScrips line 115.

Stack:
  [/opt/rt4/share/html/Helpers/PreviewScrips:115]
  [/opt/rt4/share/html/Helpers/autohandler:51]
  [/opt/rt4/sbin/../lib/RT/Interface/Web.pm:696]
  [/opt/rt4/sbin/../lib/RT/Interface/Web.pm:375]
  [/opt/rt4/share/html/autohandler:53] (/opt/rt4/sbin/../lib/RT/Interface/Web/Handler.pm:208)

### DIFF
--- a/share/html/Helpers/PreviewScrips
+++ b/share/html/Helpers/PreviewScrips
@@ -112,7 +112,7 @@ $submitted{$_} = 1 for split /,/, $ARGS{TxnRecipients};
 %                 $recips{$addr->address}++;
                   <b><%loc($type)%></b>:
 %                 my $show_checkbox = 1;
-%                 if ( grep {$_ eq $addr->address} @{$action->{NoSquelch}{$type}} ) {
+%                 if ( grep {$_ eq "$addr->address"} @{$action->{NoSquelch}{$type}} ) {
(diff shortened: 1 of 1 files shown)


cbrandtbuffalo:
It looks like this ended up being fixed with a a few other commits including https://github.com/bestpractical/rt/commit/20a9c093c8d69e575695c54f7050f18b96eaa2fe, so the core issue was solved.

PR #232, opened 2017-08-30
Path should not use loc()

mkosmach:


cbrandtbuffalo:
The code for this moved, but the issue was still there. A modified version was merged into 5.0-trunk. Thanks!

PR #233, opened 2017-09-17
Re-order the POD content

puck:
The Author and "See Also" sections were in the middle of the function
docs, move them to the end of the file.


alexmv:
Thanks for the PR -- I opted to simply those those sections entirely, since they're not really useful.

Thanks again!

PR #234, opened 2017-10-25
_FillInTemplateURL(): check for empty CF

zaucker:
The original code generates two warnings "Use of uninitialized value $value" for empty custom fields.
The same fix is applied for placeholders a few lines below.

### DIFF
--- a/lib/RT/ObjectCustomFieldValue.pm
+++ b/lib/RT/ObjectCustomFieldValue.pm
@@ -343,6 +343,7 @@ sub _FillInTemplateURL {
     # special case, whole value should be an URL
     if ( $url =~ /^__CustomField__/ ) {
         my $value = $self->Content;
+        $value = '' if !defined($value);
         # protect from potentially malicious URLs
         if ( $value =~ /^\s*(?:javascript|data):/i ) {
             my $object = $self->Object;


alexmv:
Thanks for the PR!  I swapped this to being on 4.4-trunk, and took the opportunity to switch to the new syntax which come with 5.10.

Closes #234 via 8713a66cb8.

alexmv:
Looks like I forgot to put that in the commit message of 648a76ac2b, so I _think_ you'll need to close this yourself.

zaucker:
Of course the Perl 5.10 syntax is much nicer. I just hadn't realized that I could put in "use 5.010;" myself ...
I am closing the PR as requested. Thanks for putting it into 4.4-trunk, this way we can get rid of our local patch with the next version.

zaucker:
Hmm, when I closed it, I got "unmerged commits". Please check if this was doesn't need to be merged into master as well.

alexmv:
Yeah, that's expected.  We periodically merge up `4.4-trunk` to `master`.  So this commit indeed isn't on `master` (nor on `4.4-trunk`, technically, since I rejiggered it) but it will end up so eventually.

zaucker:
Thanks for the explanation, so I'll finally close this PR.

PR #236, opened 2017-11-05
Set proper HTTP Status codes on Abort

rspier:
Initially implemented because Google's crawler was complaining that "Ticket not found" pages did not have a 404 status code.

This doesn't set status codes for all Aborts, but does the common ones.  If there's consensus as to what it should be (500?) there could be a default status on all Abort calls.

### DIFF
--- a/lib/RT/Interface/Web.pm
+++ b/lib/RT/Interface/Web.pm
@@ -72,6 +72,7 @@ use Digest::MD5 ();
 use List::MoreUtils qw();
 use JSON qw();
 use Plack::Util;
+use HTTP::Status qw();
 
 =head2 SquishedCSS $style
 
@@ -2038,6 +2039,10 @@ sub Abort {
     my $why  = shift;
     my %args = @_;
 
+    $args{Code} //= HTTP::Status::HTTP_OK;
+
+    $r->headers_out->{'Status'} = $args{Code} . ' ' . HTTP::Status::status_message($args{Code});
+
     if (   $session{'ErrorDocument'}
         && $session{'ErrorDocumentType'} )
     {
@@ -2140,11 +2145,11 @@ sub CreateTicket {
 
     my $Queue = RT::Queue->new( $current_user );
     unless ( $Queue->Load( $ARGS{'Queue'} ) ) {
-        Abort('Queue not found');
+        Abort('Queue not found', Code => HTTP::Status::HTTP_NOT_FOUND);
     }
 
     unless ( $Queue->CurrentUserHasRight('CreateTicket') ) {
-        Abort('You have no permission to create tickets in that queue.');
+        Abort('You have no permission to create tickets in that queue.', Code => HTTP::Status::HTTP_FORBIDDEN);
     }
 
     my $due;
@@ -2257,7 +2262,7 @@ sub CreateTicket {
 
     push( @Actions, split( "\n", $ErrMsg ) );
     unless ( $Ticket->CurrentUserHasRight('ShowTicket') ) {
-        Abort( "No permission to view newly created ticket #" . $Ticket->id . "." );
+        Abort( "No permission to view newly created ticket #" . $Ticket->id . ".", Code => HTTP::Status::HTTP_FORBIDDEN );
     }
     return ( $Ticket, @Actions );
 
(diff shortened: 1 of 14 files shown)


lib/RT/Interface/Web.pm
```diff
@@ -2038,6 +2039,8 @@ sub Abort {
     my $why  = shift;
     my %args = @_;
```
alexmv: This wants a default `Code` of 500 or somesuch, soas to not break the other callsites that didn't pick up a `Code`, and any non-core callsites that may call `Abort`.

rspier:
There are a lot more tests than there used to be!    I'll work through the failing ones and get you an updated PR at some point.

rspier:
Tests pass now.

Default error return code is 400, saving 500 for actual unexpected errors.  Eventually, all Abort() calls will hopefully grow a Code parameter.  Not sure if that's going to be in this PR.

Do you prefer

    Code => 404

or

    Code => HTTP::Status::HTTP_NOT_FOUND

?

Thanks!

rspier:
After some reflection, I'm wondering if we should leave the Abort() default status as 200 and override it where it makes sense.  (404, 403)  It's doing dual duty for both "user errors" and "low level errors".

There's also the issue of some browsers (IE?) showing a custom error page instead of what the server sends.  At least that used to be a thing.

rspier:
Ping?

alexmv:
Sorry for the delay!

I think sticking with default status of 200 is prooooably best, and only overriding it when there's a low-level error it makes sense to bubble up.  That's minimally invasive for the stable series, at least.

Using the `HTTP::Status::` constants is nice, since it makes it slightly more readable to those who haven't memorized status codes, and harder to typo them.

FWIW, IE showing a "friendly error page" instead of what is served [only happens on responses of < 512 bytes](https://blogs.msdn.microsoft.com/ieinternals/2010/08/18/friendly-http-error-pages/) -- [pre-gunzip'ing, of course](https://core.trac.wordpress.org/ticket/8942).

rspier:
Thanks!  Updated, tests still pass for me.  Default back to 200.  Take a look?

share/html/Ticket/ShowEmailRecord.html
```diff
@@ -105,7 +105,7 @@
 if ( not $AttachmentObj->id
     or not $AttachmentObj->TransactionId() == $Transaction ) {
     $title = loc("Error loading attachment");
-    $error = loc("Attachment '[_1]' could not be loaded", $Attachment);
+    $error = loc("Attachment '[_1]' could not be loaded", $Attachment)
```
alexmv: Unintentional change?

share/html/Dashboards/Modify.html
```diff
@@ -151,7 +151,9 @@
 if (!$Create && !$tried_create && $id && $ARGS{'Delete'}) {
     my ($ok, $msg) = $Dashboard->Delete();
     if (!$ok) {
-             Abort(loc("Couldn't delete dashboard [_1]: [_2]", $id, $msg));
+             my $code = HTTP::Status::HTTP_BAD_REQUEST;
+             $code = HTTP::Status::HTTP_FORBIDDEN if $msg =~ /Permission/;
```
alexmv: `$msg` is a translated string.  Which means you'll only be able to get 403's if your language happens to be English.

Maybe:
```
    if (!$Dashboard->CurrentUserCanDelete()) {
        Abort(loc("Couldn't delete dashboard [_1]: Permission denied", $id), Code => HTTP::Status::HTTP_FORBIDDEN);
    }
    my ($ok, $msg) = $Dashboard->Delete();
    if (!$ok) {
        Abort(loc("Couldn't delete dashboard [_1]: [_2]", $id, $msg), Code => HTTP::Status::HTTP_BAD_REQUEST);
    }
```

rspier: Done.

share/html/Ticket/ShowEmailRecord.html
rspier: Yes. Fixed.

rspier:
Thanks!  Sorry for the delayed followup.  I think I dealt with all your comments.  Tests still pass.

rspier:
You're welcome!   It was fun to hack on RT again.  It's been a long time.

alexmv:
It failed two tests for me, so it took a tiny bit more of a fixup than I thought.  Re-jiggered the first commit to work when Net::LDAP _did_ exist, and fixed up the capitalization on `share/html/Dashboards/Modify.html` in the second such that `t/web/dashboards-basics.t` passed for me.

PR #237, opened 2017-12-08
Fix acl and indexes upgrades

stbuehler:
- improve error reporting for `RT::Handle->InsertACL`
- use absolute paths for `InsertACL` and `InsertIndexes` in rt-setup-database

  Since perl 5.25.7 '.' might not be present in `@INC` by default, see: https://metacpan.org/pod/release/EXODIST/perl-5.25.7/pod/perldelta.pod#'.'-and-@INC

Also see https://forum.bestpractical.com/t/upgrading-through-rt4-1-1-fails-with-couldnt-load-acls/31855

### DIFF
--- a/lib/RT/Handle.pm
+++ b/lib/RT/Handle.pm
@@ -468,7 +468,7 @@ sub InsertACL {
     }
 
     local *acl;
-    do $path || return (0, "Couldn't load ACLs: " . $@);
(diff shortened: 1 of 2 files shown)


cbrandtbuffalo:
We ended up using a similar fix here: https://github.com/bestpractical/rt/commit/93f529acf43

Thanks for the PR!

PR #239, opened 2017-12-08
git ignore /sbin/rt-passwd

stbuehler:
`/sbin/rt-passwd` is a generated file and should be ignored.


cbrandtbuffalo:
We ran into this one also and updated with a fix in https://github.com/bestpractical/rt/commit/5c95c2d72a5.

PR #240, opened 2017-12-08
ignore missing dependencies listed in RT_IGNORE_DEPENDENCIES

stbuehler:
I need to ignore the `Mozilla::CA` dependency; on debian `LWP::Protocol::https` [doesn't use it][1], and debian [isn't going to package `Mozilla::CA`][2].

With `RT_IGNORE_DEPENDENCIES` I can simply run

```shell
export RT_IGNORE_DEPENDENCIES=Mozilla::CA
```

before `make install` and friends without having to modify the source.

[1]: https://anonscm.debian.org/cgit/pkg-perl/packages/liblwp-protocol-https-perl.git/tree/debian/patches/cert.patch
[2]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=702124

### DIFF
--- a/sbin/rt-test-dependencies.in
+++ b/sbin/rt-test-dependencies.in
@@ -109,6 +109,8 @@ $args{$_} = $default{$_} foreach grep {!exists $args{$_}} keys %default;
 $args{'with-EXTERNALAUTH-TESTS'}
     = $args{'with-EXTERNALAUTH'} && $args{'with-DEVELOPER'};
 
+my %IGNORE_DEPS = map { chomp; $_ => 1 } split(',', ($ENV{'RT_IGNORE_DEPENDENCIES'} || ''));
+
 $deps{'CORE'} = [ text_to_hash( << '.') ];
 Apache::Session 1.53
 Business::Hours
@@ -429,6 +431,7 @@ sub test_deps {
     while (@deps) {
         my $module  = shift @deps;
         my $version = shift @deps;
+        next if $IGNORE_DEPS{$module};
         my ( $test, $error ) = test_dep( $module, $version, $AVOID{$module} );
         my $msg = $module . ( $version && !$error ? " >= $version" : '' );
         print_found( $msg, $test, $error );


cbrandtbuffalo:
@stbuehler It's clearly been a while, so I'm wondering if this is still an issue? Or maybe another workaround was used?

stbuehler:
Indeed, it has been a while, I quit the job I needed this for in 2018... :)

Anyway: debian finally added [`libmozilla-ca-perl`](https://tracker.debian.org/pkg/libmozilla-ca-perl) (packaging `Mozilla::CA`) in trixie, so this specific problem should be gone now. (They patch it to use the debian standard certificates, not the mozilla ones, as they should.)

I still think it is a bad concept to list `Mozilla::CA` as direct dependency when only some other module actually uses it (and there isn't even a comment documenting that); the proper way would be `LWP::Protocol::https` with an optional "mozilla-root-cas" feature, or, because CPAN doesn't allow depending on optional features of other module (making this a rather useless concept...), a separate `LWP::Protocol::https::mozilla` package that depends on both (as suggested by https://metacpan.org/pod/CPAN::Meta::Spec#optional_features).

There might be other dependencies now (and other dists that might not need and not provide `Mozilla::CA`) that could benefit from some kind of "just ignore that dependency, I know what I'm doing"-option, but your comment suggests you'd only care if someone has that specific problem (which is fair), so I'm gonna close this.

Just for completeness: the patch removing the need for `Mozilla::CA` in `LWP::Protocol::https` on debian is still present, but now on salsa: https://salsa.debian.org/perl-team/modules/packages/liblwp-protocol-https-perl/-/blob/master/debian/patches/cert.patch

cbrandtbuffalo:
Thanks for the extra context. I dug a bit too and Mozilla::CA was originally added for rt-mailgate, so email handling. That had its own configure flag for a while to enable "ssl mailgate", so at that point it was possible to avoid it with a different option to configure. But later that was dropped and SSL mailgate became default. I'm not 100% sure why that was done. It might be because trying to change it later was hard because of the dependencies, so it was decided to default it enabled.

PR #242, opened 2018-01-24
Improve Shredder attachments regex

federicobaldo:
In order to match also _ and - characters.


sunnavy:
thanks for the commit.

"-" part looks good to me.

note that "\w" already matches "_", so it's not necessary to explicitly add "_".

federicobaldo:
Thanks, applied your suggestion.

sunnavy:
To make the merge more clear, maybe you can combine the 2 commits instead?
thanks!

federicobaldo:
Done, cheers!

PR #246, opened 2018-07-03
rt-test-dependencies should check for Locale::PO in CORE section

dsieborger:
Since commit 33fcc338b871aa95ac1d5db8da738a498ad49ef8, lib/RT/I18N/Extract.pm has required Locale::PO.  rt-test-dependencies (or, at least, the version in the stable, 4.2, 4.4 branches) doesn't check that Locale::PO is installed.  This patch moves Locale::PO from the DEVELOPER section to the CORE section.


sunnavy:
RT::I18N::Extract is actually also for dev usage only, and we made another fix(d044c162) to avoid requiring Extract.pm for RT::I18N, which has been included in 4.4-trunk and 4.4.3 release.

Thanks for reporting this.

PR #250, opened 2018-08-10
Add support to set default value(s) at CustomField creation

elacour:


cbrandtbuffalo:

PR #251, opened 2018-08-10
Allow to get only direct groups a user is member of

elacour:
Backward compatibility is assured with a default of Recursive => 1


cbrandtbuffalo:
We ended up adding the same thing in: https://github.com/bestpractical/rt/commit/0d0e862f946

We should have used this PR :) Thanks!

PR #269, opened 2018-09-06
Show Attachments: Easier UI And Real Order Of Items

mxhash:
screen shot 2018-09-06 at 14 18 05

* Adds information to the button
* Keep items in the order they had before


mxhash:
Sorry, had the wrong base for the PR. Let me know if I should recreate.

PR #277, opened 2019-02-01
Fix table wrapping error in Ticket/Update.html

mzagrabe:
Commit 3b48f214f5a3d changed the component parameter from InTable to
AsTable for calling /Ticket/Elements/EditTransactionCustomFields.
Omitting InTable causes the called component to insert an unnecessary
(and wrong) wrapping <table> element to the already existing table.
The wrapping <table> element breaks the HTML validity.


sunnavy:
Merged to 4.4-trunk, thanks!

PR #280, opened 2019-04-29
Fix trouble with udefined Lifecycle type

nicqrocks:
I was getting a problem on my system with calls to `Lifecycle->Load` having a undefined type value, which caused the ticket page not to load. I tracked the problem down to this file. Everywhere else seems to give `Type` a default value of "ticket", except for this element. I changed this on my system and was able to get pages to load again.

Granted, this error showed up after progressively upgrading the system from 4.4.1 to 4.4.4, so if no one else has trouble with this, then this change could probably be ignored.


sunnavy:
I merged a branch that has the same fix, thanks for your report!

PR #282, opened 2019-07-25
Init callback for self service create.html

mzagrabe:
Allow SelfService/Create.html the same courtesy as Ticket/Create.html for modifications after <%init> block.


cbrandtbuffalo:
Callback added for RT 4.4 in c25ba3724.

PR #284, opened 2019-09-09
Add --update-privilege flag to rt-ldapimport

vinzBad:
I needed to update the Privilege status of many already imported users.


vinzBad:
any feedback?

PR #285, opened 2019-10-05
Allow login with any field from attr_match_list

paierh:
The current implementation of external users does not allow users to use multiple attributes (e.g. uid and email address) to logon to RT. With this simple modification any attribute could be used to identify a user as long as it results in an unique identification.
 
The UserExists method now searches in the LDAP directory for a match in any of the fields specified with attr_match_list. After an unique match is found (e.g. uid or email address) the provided username is updated to always reflect the value of the "Name" field. Now a user can use any element specified in attr_match_list to logon as long as the provided login data results in an unique result.

I'm not sure if you approve of by-ref-changing of method parameters. An alternative approach would be to rewrite UserExists so it returns the actual username as a string after looking for a match in LDAP or in a database or an empty string if no (unique) match is found.

### DIFF
--- a/lib/RT/Authen/ExternalAuth.pm
+++ b/lib/RT/Authen/ExternalAuth.pm
@@ -661,6 +661,9 @@ sub UserExists {
     } elsif ($config->{'type'} eq 'ldap') {
         $success = RT::Authen::ExternalAuth::LDAP::UserExists($username,$service);
     }
+    if ($success) {
+        $_[0] = $username;
+    }
 
     return $success;
 }
--- a/lib/RT/Authen/ExternalAuth/LDAP.pm
+++ b/lib/RT/Authen/ExternalAuth/LDAP.pm
@@ -537,14 +537,21 @@ sub UserExists {
     undef $filter if defined $filter and $filter eq "()";
 
     if (defined($config->{'attr_map'}->{'Name'})) {
+        my $attr_map        = $config->{'attr_map'};
         # Construct the complex filter
+        my $filter_addition = '(|';
+        foreach my $rt_attr (@{$config->{'attr_match_list'}}) {
+            $filter_addition .= '(' .
+                                $attr_map->{$rt_attr} .
+                                '=' .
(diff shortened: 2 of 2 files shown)


cbrandtbuffalo:
I believe attr_match_list already supports trying multiple attributes for login. It currently attempts the search one at a time. Looking at the commit, it seems the main change is to instead do a single search query with an or (|) instead. Is that correct?

paierh:
Actually the main change is to canonicalize $username so that there is only one user in the database for each external user. 
At the moment we have the following situation: 
For example let's say that my username is paierh and my email address is hp@rt-server.net. 
On my fisrt login using paierh as username a new user object is created in the RT database with Name=paierh and email=hp@rt-server.net 
On my second login using hp@rt-server.net another NEW user obeject is created with Name=hp@ rt-server.net and Email=NULL instead of loading the existing user from the database. 

My patch allows external authentication providers to canonicalize the provided login name before it is looked up in the database. 

I solved this by allowing ExternalAuth:: LDAP::UserExists to update the 1st parameter ($username) with the LDAP-Value of attr_name->Name using: 
$_ [0] = ( $entry -> get_value( $key ))[0]; 

And then did the same to allow ExternalAuth::UserExists to update $username using: 
$_ [0] = $username ; 

I did not touch DB::UserExists, but the same change would work there also. 

I choose the "by ref update" of $_[0] to avoid changes to the function interface, but I think a better way to solve the problem would one of the following: 


    * Change the return value of UserExists from boolean to string with the canonicalized username as a result 
    * Add an additional function call to DoAuth before $session -> { ' CurrentUser ' } -> Load( $username ); to call a new function CanonicalizeUserName 
Regards 
harald

sunnavy:
Hi harald

We recently updated related code, to support login using either name or email address, and I believe it could help the scenario of "only one user in the database for each external user".

Could you confirm? Note that you need to specify both "Name" and "EmailAddress" in attr_match_list.

Thanks!
-sunnavy

PR #287, opened 2019-10-17
Prevent users to enter values below 120s

tbrumm:
Prevent users to enter directly values below a given time.

Many user changes the HomeRefreshInterval via URL directly to very short values.


elacour:
Nice idea, thought I would suggest using lower value of @RefreshIntervals rather than hardcoding 120. For example I have a pretty busy RT instance where the lowest value of RefreshIntervals is 300, not 120.

tbrumm:
You are right. Possibly reading the Config Settings for this would be much better. This was a first working draft for us.

PR #290, opened 2019-12-06
support multi-word CustomRole names in Notify Action

korylprince:
Currently we can create a CustomRole with names that include spaces, e.g. `Queue Manager`.

However, the `Notify.pm` Action only supports CustomRole names that match the regex `( \w+)`, i.e. a single word.

This pull request modifies the regex to allow for multiple words in the CustomRole name.

### DIFF
--- a/lib/RT/Action/Notify.pm
+++ b/lib/RT/Action/Notify.pm
@@ -113,7 +113,7 @@ sub SetRecipients {
                            # then RT::CustomRole-# or a role name
                            (?:
                                RT::CustomRole-(\d+)    # $2 role id
-                             | ( \w+ )                 # $3 role name
+                             | ( (?: \w\s?)+ )         # $3 role name
                            )
 
                            # then, optionally, a type after a slash


cbrandtbuffalo:
Hi @korylprince , thanks for your PR. We have a branch that also has a fix for this issue: https://github.com/bestpractical/rt/tree/4.4/parse-notify-argument
Would you be willing to test that out? You should need just one commit as the other two are tests:
https://github.com/bestpractical/rt/commit/1e28eef1981ae4350eefbf31c0348cf6a498dfab

korylprince:
@cbrandtbuffalo, thanks, I didn't see that branch.

I've tested it out and it seems to solve this use case. It looks like it's been around awhile. Do you know when it will be merged?

cbrandtbuffalo:
> I've tested it out and it seems to solve this use case. It looks like it's been around awhile. Do you know when it will be merged?

Great, thanks for testing and letting us know. I don't have an exact date, but we plan to include this in 4.4.5, so it should be in the next release. I'll close this, but feel free to comment if you have more questions.

PR #293, opened 2020-03-06
Create go.yml

ghaniqiq:


cbrandtbuffalo:
The purpose of this new empty file isn't clear from the commit message.

PR #295, opened 2020-04-07
Leave umask flag

thirdwheel:
I was guiding someone on setting up a web based application using OpenBSD's httpd when I noticed that the unix domain socket RT was using for httpd to communicate via FastCGI was world-writable.  I didn't think this was very secure and looked into why it was doing that, since I hadn't explicitly told RT to do that.  I finally tracked it down to an option sent to Plack's FCGI handler that told the handler whether to set the umask to 0 (effectively make the file world-writable) or use the already available umask.

It turned out to be very simple to pass this flag through to Plack's FCGI handler.  I added some documentation to the RT-Config.pm.in file to make clear what was going on and that you should ensure you have the right umask that would allow the web server to communicate with RT.

### DIFF
--- a/etc/RT_Config.pm.in
+++ b/etc/RT_Config.pm.in
@@ -149,6 +149,21 @@ Set( @StaticRoots, () );
 
 =back
 
+=item C<$LeaveUMask>
+
+Set C<$LeaveUMask> to 1 to tell the Plack runner to preserve the umask
+of the running process.  By default, Plack will set the umask to 0,
+essentially making the socket file world-writeable.  
+
+This setting only applies if you are using FastCGI on UNIX domain 
+sockets.
+
+=cut
+
+Set( @$LeaveUMask, 0 );
+
+=back
+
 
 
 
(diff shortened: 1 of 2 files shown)


sunnavy:
Hi Aaron

Thanks for reporting this. I'm curious of  the way you set up the RT web server, maybe it's Apache+mod_fastcgi?

Regards
sunnavy

thirdwheel:
No, it's RT running as a FastCGI daemon and OpenBSD's httpd set to direct all requests to it via a unix domain socket.

I have been documenting the setup process on a repo set up for OpenBSD's httpd by its maintainer: https://github.com/reyk/httpd/wiki/Running-Request-Tracker-(RT)-with-OpenBSD-httpd

sunnavy:
Instead of adding a new config, it's probably more flexible to set it from CLI, e.g.

    /opt/rt4/sbin/rt-server.fcgi --leave-umask 1 --socket $RT_SOCK ...

How do you think?

Regards
sunnavy

thirdwheel:
Yep, that works.  I'll close this off now, thanks!

PR #296, opened 2020-04-20
bugfix: Correctly assign both return values of $Ticket->CurrentUserCanSetOwner()

cure:
This applies to share/html/Elements/Tabs and share/html/Ticket/Elements/ShowSummary.

This fixes a bug in the ticket menu tab for users without rights to set the current ticket owner: the People tab will no longer be shown in that case.

Similarly, the 'People' tab under 'Ticket Metadata' on the ticket summary page will no longer be clickable in that case.

This bug was introduced in commit 43f0e872019ca3abd20168730905477dac98785d


sunnavy:
Hi cure

Thanks for the PR. This issue has been fixed in 5.0 and I ported it back to 4.2/4.4 accordingly just now.

Regards
sunnavy

PR #298, opened 2020-09-09
upgrades: fix execution of ACL upgrades

rissson:
During the ugprade of our RT from 4.4.2 to 5.0.0, we ran through a couple of issues. Here are the fixes we had to make in order to make the upgrade scripts work.

Cc. @Byh0ki

### DIFF
--- a/etc/acl.Pg
+++ b/etc/acl.Pg
@@ -10,41 +10,41 @@ sub acl {
         Attributes
         attributes_id_seq
         queues_id_seq
-        Queues 
+        Queues
         links_id_seq
-        Links 
+        Links
         principals_id_seq
-        Principals 
+        Principals
         groups_id_seq
-        Groups 
+        Groups
         scripconditions_id_seq
-        ScripConditions 
+        ScripConditions
         transactions_id_seq
-        Transactions 
+        Transactions
         scrips_id_seq
-        Scrips 
+        Scrips
         objectscrips_id_seq
         ObjectScrips
         acl_id_seq
-        ACL 
+        ACL
         groupmembers_id_seq
-        GroupMembers 
+        GroupMembers
         cachedgroupmembers_id_seq
-        CachedGroupMembers 
+        CachedGroupMembers
         users_id_seq
-        Users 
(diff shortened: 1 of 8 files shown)


sunnavy:
I believe the issue has been fixed in 4.4/upgrade-acl-abs-path, could you check if it works for you?

Thanks!

rissson:
> I believe the issue has been fixed in 4.4/upgrade-acl-abs-path, could you check if it works for you?

This only fixes one of the issues we encountered. Why hasn't this landed in the tarball that can be downloaded from the 5.0.0 release tag?

sunnavy:
Could you explain the other issue? Correct me if I misunderstood, but the other commit(136455b357)  just cleans spaces, adds optional trialing ";", etc.

FYI, 4.4/upgrade-acl-abs-path or a similar fix will be included in the next release.

rissson:
> Could you explain the other issue? Correct me if I misunderstood, but the other commit([136455b](https://github.com/bestpractical/rt/commit/136455b357396e84e51380572cb7367989c66ecb)) just cleans spaces, adds optional trialing ";", etc.

We were getting an error from [this](https://github.com/bestpractical/rt/blob/stable/lib/RT/Handle.pm#L467) line. Adding the optional trailing `;` resolved the issue. I have no idea why but they did.

> FYI, 4.4/upgrade-acl-abs-path or a similar fix will be included in the next release.

Understood. It's a shame it didn't make it for the 5.0.0, some people my be experiencing the same issue as we did.

sunnavy:
4.4/upgrade-acl-abs-path fixes the error you got from https://github.com/bestpractical/rt/blob/stable/lib/RT/Handle.pm#L467, which you fixed in the other commit(c75781fb91) in a different way, I don't think it's related to any optional trailing ";" or space.

rissson:
Okay, we'll try the changes made in `4.4/upgrade-acl-abs-path` and close this if they work, thanks.

rissson:
I'm not sure we'll ever have time to try this so I'm closing it for now.

PR #303, opened 2021-02-19
Add webpath to href in RelatedData 

jtlarson:
Issue: href's in RelatedData are missing RT webpath variable. Using the buttons in RT results in 404 error if "WebPath" is defined in RT_SiteConfig.pm

Fix tested in 5.0.0


sunnavy:
Merged and thank you!

-sunnavy

PR #304, opened 2021-03-05
Do not remove WebPath from undefined result page

voegelas:
This commit fixes the bug reported in [I#37068](https://rt.bestpractical.com/Ticket/Display.html?id=37068). RT outputs the below warning when $WebPath is set in the configuration.

    Use of uninitialized value $result_page in substitution (s///) at lib/RT/Interface/Web/MenuBuilder.pm line 704.


sunnavy:
Merged and thank you!

-sunnavy

PR #308, opened 2021-05-10
Priority class cleanup updates

elacour:
When PriorityAsString was shipped as an RT extension it was needed to duplicate the CSSClass method to be compatible with RT 4.0. It's no longer needed now that it has been cored in RT 5.
Also there was two places where clean up of spaces/non-ascii chars was missing.


sunnavy:
Merged and thanks!

-sunnavy

PR #313, opened 2021-10-09
GROUPS was added to the list of reserved words in MySQL v8.0.2

Dudley5000:
Backticks are necessary here to enable the reserved word to be treated as an identifier, otherwise installation fails at the point where the `Groups` table is created.

See https://dev.mysql.com/doc/refman/8.0/en/identifiers.html

### DIFF
--- a/etc/schema.mysql
+++ b/etc/schema.mysql
@@ -70,7 +70,7 @@ CREATE TABLE Principals (
 ) ENGINE=InnoDB CHARACTER SET utf8mb4;
 
 
-CREATE TABLE Groups (
+CREATE TABLE `Groups` (
   id INTEGER NOT NULL  AUTO_INCREMENT,
   Name varchar(200) NULL  ,
   Description varchar(255) NULL  ,
@@ -83,8 +83,8 @@ CREATE TABLE Groups (
   PRIMARY KEY (id)
 ) ENGINE=InnoDB CHARACTER SET utf8mb4;
 
-CREATE INDEX Groups1 ON Groups (Domain, Name, Instance);
-CREATE INDEX Groups2 On Groups (Instance);
+CREATE INDEX Groups1 ON `Groups` (Domain, Name, Instance);
+CREATE INDEX Groups2 On `Groups` (Instance);
 
 CREATE TABLE ScripConditions (
   id INTEGER NOT NULL  AUTO_INCREMENT,


Dudley5000:
I probably should have read the `README.MySQL`

> Note that MySQL 8 is not yet supported because of changes to the group keyword.

joelpittet:
FYI I got it working with this on an existing 4.4.4 RT site:
https://forum.bestpractical.com/t/temporary-solution-for-missing-mysql-8-support/34568/5

And looks like there is a 4.x branch that someone tried going at this:
https://github.com/bestpractical/rt/tree/4.4/mysql8-quoted-tables

cbrandtbuffalo:
> FYI I got it working with this on an existing 4.4.4 RT site: https://forum.bestpractical.com/t/temporary-solution-for-missing-mysql-8-support/34568/5
> 
> And looks like there is a 4.x branch that someone tried going at this: https://github.com/bestpractical/rt/tree/4.4/mysql8-quoted-tables

Updates to run with MySQL 8 have been merged for the 4.4 and 5.0 series and will be in the next release. We also updated our github actions, so all branches are currently running our automated tests against MySQL 8. Note that you'll also need to update the module DBIx::SearchBuilder.

joelpittet:
@cbrandtbuffalo that is awesome news! Thanks, hopefully we'll update to 5.0 sooner than later but appreciate you've got it working in 4.4 as well.

I was learning Perl for this and I wondered how module dependencies are managed (or how you manage them) We've upgraded manually `DBIx::SearchBuilder` as mentioned in the forum. That was nice because it cleaned up the messy ```Sub Table { '`Groups`' }``` that I needed to hack around to get it to work prior to that 1.74 version.

cbrandtbuffalo:
> I was learning Perl for this and I wondered how module dependencies are managed (or how you manage them)

RT runs a module dependency check as part of installing or upgrading and it also knows about module versions. It's easier to see the dependency file in RT 5:

https://github.com/bestpractical/rt/blob/stable/etc/cpanfile

That is used by this script and it prompts you to update modules if a newer version is needed:

https://github.com/bestpractical/rt/blob/stable/sbin/rt-test-dependencies.in

joelpittet:
@cbrandtbuffalo Should https://github.com/bestpractical/rt/blob/stable/etc/cpanfile#L25 be `1.74` then for MySQL 8 support?

Or I guess on the trunk branch? https://github.com/bestpractical/rt/blob/5.0-trunk/etc/cpanfile#L25

cbrandtbuffalo:
> Should https://github.com/bestpractical/rt/blob/stable/etc/cpanfile#L25 be `1.74` then for MySQL 8 support?

I think that version is needed for a different new feature, but yes, that will be sufficient for MySQL 8 support. Usually when you update a module during an upgrade, you'll get the latest available version anyway.

PR #321, opened 2021-12-30
Make setting sqldump actually work

grifferz:
If `--sqldump` is set then the filepath will end up in `$tmp`. The later call to `GetOptions( 'no-sqldump' => \$tmp )` will always succeed because it is a valid getopts specification (it just doesn't set anything). Then the `&& $tmp` condition will also pass because `$tmp` still has a value from when `--sqldump` earlier set it. The result is that `$opt{'no-sqldump'}` will end up being set to the filepath of the dump. Later on, this will seem like `--no-dump` was actually requested and no SQL dump will take place.

Fix this by unsetting `$tmp` before it is tested again.

### DIFF
--- a/sbin/rt-shredder.in
+++ b/sbin/rt-shredder.in
@@ -253,6 +253,7 @@ sub parse_args
     if( GetOptions( 'sqldump=s' => \$tmp ) && $tmp ) {
         $opt{'sqldump'} = $tmp;
     }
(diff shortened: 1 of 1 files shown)


sunnavy:
Thanks, merged into 4.0-trunk, which will be included in the next versions of RT 4 and RT 5.

-sunnavy

PR #324, opened 2022-04-12
ExtractSubjectTag Action: prepend, not append, remote system's subject tag

Al2Klimov:
Too long (200+) ticket subjects (produced by this action) are shortened automatically.
This could cut off the remote system's subject tag if at the end.

### DIFF
--- a/lib/RT/Action/ExtractSubjectTag.pm
+++ b/lib/RT/Action/ExtractSubjectTag.pm
@@ -122,7 +122,7 @@ sub Commit {
                 next TAGLIST;
             }
         }
-        $TicketSubject .= " $tag" unless ( $TicketSubject =~ /\Q$tag\E/ );
+        $TicketSubject = "$tag $TicketSubject" unless ( $TicketSubject =~ /\Q$tag\E/ );
     }
 
     $self->TicketObj->SetSubject($TicketSubject)


Al2Klimov:
## How to reproduce

1. Setup two RTs, say, _Tom_ and _Jerry_
2. In _Tom_ create a ticket with a _loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong_ subject, add _Jerry_ as requestor and reply something
3. In _Jerry_ reply to the ticket
4. In _Tom_ reply to the ticket -> opens separate ticket at _Jerry_

Al2Klimov:
## Why exactly this happens

To understand the following we need to remember that there are _ticket subjects_ and _eMail subjects_.

1. In _Tom_ the newly created ticket has the ID 1 and the _ticket subject_ `looooooo...` (cut off).
2. On "reply" (it's not actually a reply, but the very first message of the whole conversation) _Tom_ sends an eMail to _Jerry_ with the _eMail subject_ `[%s #%d] %s`, i.e. `[tom #1] looooooo...` (cut off).
3. That eMail creates a new ticket in _Jerry_ with ID 1 and the _eMail subject_ from above as _ticket subject_.
4. On (actual) reply _Jerry_ sends an eMail to _Tom_ with the _eMail subject_ `[%s #%d] %s`, i.e. `[jerry #1] [tom #1] looooooo...` (cut off).
5. _Tom_ recognises `[tom #1]` in the _eMail subject_ and associates the eMail with the existing ticket.
6. _Tom_ also recognises the foreign tag `[jerry #1]` and the code I changed in this PR tries(!) to change the _ticket subject_ to `loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong [jerry #1]` (so that on next reply from _Tom_ gets into the right ticket at _Jerry_). It tries! But, as master Yoda said: Do or do not. There's no _try_. Effectively _Tom_ doesn’t change the _ticket subject_ as above as ` [jerry #1]` gets cut off silently due to the database column length.
7. As the _ticket subject_ stays as-is the next reply from _Tom_ misses `[jerry #1]`.
8. Therefore _Jerry_ can’t associate it with the already existing ticket and creates a new one.

Al2Klimov:
## How to workaround

At least in the _Tom_ RT, preferably in both ones:

1. Navigate to Admin/Scrips/Select
2. Open _On transaction, add any tags in the transaction's subject to the ticket's subject_
3. Set _Action_ to _User Defined_
4. Set _Custom action preparation code_ to `1;`
5. Copy the code given below to _Custom action commit code_
6. Apply this PR as patch
7. _Save Changes_

https://github.com/bestpractical/rt/blob/cbfe3cd385ef5e2c7fded70367348d80c965c6d1/lib/RT/Action/ExtractSubjectTag.pm#L103-L131

CC @lippserd

PR #327, opened 2022-05-03
disable browser spell check on customcode

firefart:
This PR disables the browser spell checking feature on the CustomCode textarea in the admin field as it is not really compatible with perl code


sunnavy:
Merged and thanks!

-sunnavy

PR #328, opened 2022-05-09
Keep the Owner when using SelectOwnerAutocomplete

INCIBE-CERT:
When going through SelectOwnerAutocomplete, the load
of a template on investigation launch passes the selected
value in $Default as text, but SelectOwnerAutocomplete only
handles the field as a numeric id.
Just copying $Default to $value fixes the issue.


sunnavy:
> if ( $Default and not $Default =~ /\D/ ) {

Maybe update the former condition to handle Name there? i.e.

if ( $Default ) {

-sunnavy

INCIBE-CERT:
Hi sunnavy!

If it already knows the name, it doesn't really need to do the `$user->Load($Default);` to fetch it, but User::Load would end up doing a `$self->LoadByCol( "Name", $identifier );` so it should work as well. I have no special preference.

PR #329, opened 2022-05-10
Do not check acl when auto-setting core date fields

elacour:
When using custom rights in lifecycles, a user with only a right such as
"CloseTicket" should trigger the update of "Resolved" field without
"ModifyTicket" right.


sunnavy:
Merged and thanks!

-sunnavy

PR #332, opened 2022-05-11
All commands should have the "RT CMD BOILERPLATE"

chutzimir:
No idea why these two were missed, but rt-mailgate certainly breaks when
the system libwww-perl is out of date.

### DIFF
--- a/bin/rt-mailgate.in
+++ b/bin/rt-mailgate.in
@@ -55,6 +55,23 @@ rt-mailgate - Mail interface to RT.
 use strict;
 use warnings;
 
+# fix lib paths, some may be relative
+BEGIN { # BEGIN RT CMD BOILERPLATE
+    require File::Spec;
+    require Cwd;
+    my @libs = ("@RT_LIB_PATH@", "@LOCAL_LIB_PATH@");
+    my $bin_path;
+
+    for my $lib (@libs) {
+        unless ( File::Spec->file_name_is_absolute($lib) ) {
+            $bin_path ||= ( File::Spec->splitpath(Cwd::abs_path(__FILE__)) )[1];
+            $lib = File::Spec->catfile( $bin_path, File::Spec->updir, $lib );
+        }
+        unshift @INC, $lib;
+    }
+
+}
+
 use Getopt::Long;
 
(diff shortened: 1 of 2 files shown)


cbrandtbuffalo:
This looks to be a duplicate of #331, so closing this one.

chutzimir:
Not a duplicate, this one is to merge to `stable`, #331 is to merge to `maint`. I could not figure out which one is the right way to send this.

cbrandtbuffalo:
Ah, got it. We merge up from older to newer, so if we want a change in 4.4 and 5.0, we just add to 4.4 and it will end up in 5.0. So we'll need just one.
For this change, we have a newer method for CLI scripts. You can see an example in this commit 6076e503. In that case, we updated it only because we were maintaining documentation in two places, so it allowed us to improve that.
I'm also curious what issue you saw that was fixed with your PR? Those scripts should run as-is, but it seems like you had a dependency issue that showed up with your changes, or was if fixed somehow?

chutzimir:
> I'm also curious what issue you saw that was fixed with your PR? Those scripts should run as-is, but it seems like you had a dependency issue that showed up with your changes, or was if fixed somehow?

We used to run RT on an OS with outdated perl libraries (CentOS 6 maybe?). Since we do not want to touch the system libraries, we installed them from CPAN into RT's libdir and packaged the whole thing as an RPM. The method works great, because all the RT tools are configured to look for libraries in RT and these libraries do not interfere with anything outside.

As I recall, `rt-mailgate` did not like the old version of LWP on the system, so it needed the more recent one bundled with RT. It is probably not so relevant anymore as you pointed out.

cbrandtbuffalo:
Ah, the "installed into RT's libdir" explains it. The "RT CMD BOILERPLATE" pulls in the RT libraries even when RT is installed in different places, and if you were installing additional perl libraries there, that would have helped rt-mailgate find the right ones.

Typically we don't include RT libraries with rt-mailgate because it uses the REST interface and doesn't actually need all of the core RT code. This allows people to install rt-mailgate on an external server, like a mail server, for example.

I'm going to close out the other PR as well, but thanks for submitting!

PR #333, opened 2022-05-16
Change display from block to inline for create elements

mzagrabe:
Employing the callback Elements/CreateTicket/InFormElement to add
elements to the nav bar requires inline display for the create button
otherwise the button spills over due to block display.


sunnavy:
Merged to 5.0-trunk.

Thanks!

-sunnavy

PR #334, opened 2022-06-01
Update Condition.pm

bit0mike:
Add missing semicolon.  Perl 5.36 won't load the module without it.

That in turn causes RT to silently (silently in the web UI, not-so-silently in the logs) refuse to send any emails after updating from Perl 5.34 to 5.36.


sunnavy:
The fix is in 4.4-trunk already and will be merged up to 5.0-trunk soon.

Thanks for the PR anyway.

-sunnavy

PR #335, opened 2022-07-20
Rewind uploaded file after reading

elacour:
$fh is a ref to the filehandle given by cgi. Hence if this method is called
twice (it is if you installed RT::Extension::MandatoryOnTransition with
Upload/Image CF support[1]) the second reading will return empty data because
it starts at end of file.

[1] - https://github.com/bestpractical/rt-extension-mandatoryontransition/pull/6


sunnavy:
Merged and thanks!

-sunnavy

PR #336, opened 2022-07-21
Sessions are not saved for usernames with dots

chutzimir:
Usernames can contain periods, and many other non-whitespace characters.

Without this patch a user with a `.` in their username cannot have their
session saved to `~/.rt_sessions`.


sunnavy:
RT actually supports spaces in username too, so I created 5.0/arbitrary-user-name-in-rt-sessions for it.

Thanks!

-sunnavy

chutzimir:
Is this only going to be merged to 4.4?

cbrandtbuffalo:
> Is this only going to be merged to 4.4?

No, not only. We merge up to newer versions, so changes added to 4.4 will end up in 5.0. We typically do this as part of preparing for new releases.

chutzimir:
> > Is this only going to be merged to 4.4?
> 
> No, not only. We merge up to newer versions, so changes added to 4.4 will end up in 5.0. We typically do this as part of preparing for new releases.

Sorry, I meant to write "Is this **also** going to be merged to 4.4?" 🤣

PR #338, opened 2022-07-21
Display Richtext fields in a readable manner

chutzimir:
This makes the text readable on the ticket history page, which was
strongly requested by our users.

### DIFF
--- a/share/html/Elements/ShowTransaction
+++ b/share/html/Elements/ShowTransaction
@@ -122,7 +122,7 @@ my @classes = (
     ($RowNum % 2 ? 'odd' : 'even')
 );
 
-my $desc = $Transaction->BriefDescriptionAsHTML;
+my $desc = $Transaction->BriefDescription;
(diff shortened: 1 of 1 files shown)


sunnavy:
Hmm, as the page is html,  the html version(BriefDescriptionAsHTML)  shall be rendered fine.

Could you show me an example that the change fixes?

-sunnavy

chutzimir:
The latest version renders fine, but the ticket history is hard to read if you want to see how a value changed. I can get a screenshot in a couple of weeks if it is going to help.

chutzimir:
I made a blank ticket and only updated the "Request Information" custom field. Then I changed its content. Below is what the ticket history looks like now:

image

And below is what it looks like with this patch.

image

It is not perfect, as some styling is applied in the history message, but it is much more readable compared to looking at all those `&nbsp;` and tags that it gets littered with otherwise.

PR #340, opened 2022-07-31
A tiny tweak to the gitignore file to ignore all of var/

obra:
This one ought to be pretty straightforward - it just ignores more stuff that shouldn't be checked in.


sunnavy:
Merged accordingly.

Thanks!

-sunnavy

PR #339, opened 2022-07-30
Replace boilerplate HTML for rows in key-value lists with a standardized component

obra:
This PR adds a new "LabelValueRow" component to extract and standardize how we show key-value pairs in lists in RT's UI.

This makes it easier to keep the style consistent, easier to update the style later and reduces boilerplate html in code.

Ideally, this PR would result in zero user-visible changes within RT.

As a demo, I've updated Ticket/Elements/ShowBasics to use the new component. 

This PR is currently a *draft* for initial review. If y'all believe this would be desirable change, I'm happy to update the entire codebase to use the new component, either as one PR or as a series of smaller PRs.

Since it's going to touch a lot of the UI, if you want the updates all together, I should probably do those updates in one pass in advance of a relatively quick merge or there will be numerous merge conflicts. 

If you want me to keep going, please let me know which branch I should be targeting and whether you have stylistic preferences.

PR #341, opened 2022-08-01
extract and standardize how we show key-value pairs in lists in RT's UI.

obra:
This PR adds a new "LabelValueRow" component to extract and standardize how we show key-value pairs in lists in RT's UI.

This makes it easier to keep the style consistent, easier to update the style later and reduces boilerplate html in code.

Ideally, this PR would result in zero user-visible changes within RT.

As a first step, I've updated Ticket/Elements/ShowBasics to use the new component.

If y'all believe this would be desirable change, I'm happy to update the entire codebase to use the new component, either as one PR or as a series of smaller PRs.

Since it's going to touch a lot of the UI, if you want the updates all together, I should probably do those updates in one pass in advance of a relatively quick merge or there will be numerous merge conflicts.

If you want me to keep going, please let me know which branch I should be targeting and whether you have stylistic preferences.

This replaces #339

### DIFF
--- a/share/html/Ticket/Elements/EditPeopleInline
+++ b/share/html/Ticket/Elements/EditPeopleInline
@@ -51,11 +51,9 @@
   </div>
 </div>
 
-<div class="form-row">
-  <div class="label col-3"><&|/l&>Owner</&>:</div>
-  <div class="value col-6"><& /Elements/SelectOwner, Name => 'Owner', QueueObj => $Ticket->QueueObj, TicketObj => $Ticket, Default => $Ticket->OwnerObj->Id, DefaultValue => 0&></div>
-</div>
-
+<&| /Elements/LabeledValue, Label => loc('Owner') &>
+ <& /Elements/SelectOwner, Name => 'Owner', QueueObj => $Ticket->QueueObj, TicketObj => $Ticket, Default => $Ticket->OwnerObj->Id, DefaultValue => 0&>
+</&>
 % my @role_fields;
 % my $single_roles = $Ticket->QueueObj->CustomRoles;
 % $single_roles->LimitToSingleValue;
@@ -65,15 +63,9 @@
 % $m->callback( CustomRoles => $single_roles, SingleRoles => 1, Ticket => $Ticket, %ARGS, CallbackName => 'ModifyCustomRoles', CallbackPage => '/Ticket/ModifyPeople.html' );
 
 % while (my $role = $single_roles->Next) {
-<div class="form-row">
-  <div class="label col-3">
-    <span class="<% $role->EntryHint ? 'prev-icon-helper' : '' %>"><% $role->Name %>:</span>\
-% if ( my $hint = $role->EntryHint ) {
-<span class="far fa-question-circle icon-helper" data-toggle="tooltip" data-placement="top" data-original-title="<% $hint %>"></span>
-% }
-  </div>
-  <div class="value col-6"><& /Elements/SingleUserRoleInput, role => $role, Ticket => $Ticket &></div>
-</div>
+<&| /Elements/LabeledValue, Label => $role->Name, LabelTooltip => $role->EntryHint &>
+  <& /Elements/SingleUserRoleInput, role => $role, Ticket => $Ticket &></div>
+</&>
 
 % }
 
--- a/share/html/Admin/Actions/Elements/EditBasics
+++ b/share/html/Admin/Actions/Elements/EditBasics
@@ -45,42 +45,18 @@
 %# those contributions and any derivatives thereof.
 %#
 %# END BPS TAGGED BLOCK }}}
-<div class="form-row">
-  <div class="label col-3">
-    <&|/l&>Name</&>:
-  </div>
-  <div class="value col-9">
+<&| /Elements/LabeledValue, Label => loc("Name") &>
     <input type="text" class="form-control" name="Name" size="60" value="<% $ARGS{"Name"} || $Action->Name || '' %>" />
-  </div>
-</div>
-
-<div class="form-row">
-  <div class="label col-3">
-    <&|/l&>Description</&>:
-  </div>
-  <div class="value col-9">
+</&>
+<&| /Elements/LabeledValue, Label => loc("Description") &>
     <input type="text" class="form-control" name="Description" size="60" value="<% $ARGS{"Description"} || $Action->Description || '' %>" />
-  </div>
-</div>
-
-<div class="form-row">
-  <div class="label col-3">
-    <&|/l&>Action Module</&>:
-  </div>
-  <div class="value col-9">
+</&>
+<&| /Elements/LabeledValue, Label => loc("Action Module") &>
     <input type="text" class="form-control" name="ExecModule" size="60" value="<% $ARGS{"ExecModule"} || $Action->ExecModule || ''%>" />
-  </div>
-</div>
-
-<div class="form-row">
-  <div class="label col-3">
-    <&|/l&>Parameters to Pass</&>:
-  </div>
-  <div class="value col-9">
+</&>
+<&| /Elements/LabeledValue, Label => loc("Parameters to Pass") &>
     <input type="text" class="form-control" name="Argument" size="60" value="<% $ARGS{"Argument"} || $Action->Argument || '' %>"/>
-  </div>
-</div>
-
+</&>
 <%ARGS>
 $Action
 </%ARGS>
--- a/share/html/Admin/Actions/Elements/ShowBasics
+++ b/share/html/Admin/Actions/Elements/ShowBasics
@@ -45,41 +45,18 @@
 %# those contributions and any derivatives thereof.
 %#
 %# END BPS TAGGED BLOCK }}}
-<div class="form-row">
-  <div class="label col-3">
-    <&|/l&>Name</&>:
-  </div>
-  <div class="value col-9">
+<&| /Elements/LabeledValue, Label => loc("Name") &>
     <span class="current-value"><% $ARGS{"Name"} || $Action->Name || '' %></span>
-  </div>
-</div>
-
-<div class="form-row">
-  <div class="label col-3">
-    <&|/l&>Description</&>:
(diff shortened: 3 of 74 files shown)


cbrandtbuffalo:
Looks ok to me and makes sense. I would also add args for the col-3, col-9 classes. Default to col-3/9, but allow override. I think we have some cases with different values, like for transaction CFs. And there might be something in assets.

If you wanted to do it everywhere, I would be ok to merge since, as you noted, it should result in no visible changes.

share/html/Elements/LabelValueRow
```diff
@@ -0,0 +1,74 @@
...
+</%args>
+<%init>
+</%init>
+<div class="<% $class %> form-row">
+    <div class="label col-3"><% $label %>:</div>
+    <div class="value col-9">
+<%perl>
+     if ($value_span_class) {
+         $m->out('<span class="'.$m->interp->apply_escapes($value_span_class,"h").'">');
+     }
```
cbrandtbuffalo, line 63: Nitpick, there's a trailing space here and on line 71.

sunnavy:
I like the idea.  2 suggestions:

1. We use CamelCase argument names in most mason templates, maybe stick to that for consistency?

BTW, the 2 additional args Jim mentioned are $LabelCols and $ValueCols, which are used in other mason code like /Elements/ShowCustomFields

2. Maybe make LabelRowValue support filtering too?

With it, we won't need to create things like "ShowTimeWorkedByUser", we can insert content directly like:

    <&| /Elements/LabelRowValue, ... &>
        # the content of ShowTimeWorkedByUser
    </&>

share/html/Elements/LabelValueRow
obra: ty

obra:
I'm 100% happy to fix the stylistic issues. 

I'll look at supporting filtering (although I do think there's some value to breaking out complicated bits of rendering into components.)

I explicitly don't want to have calling code pass in hardcoded width values because those would force more brittleness into the layout engine that I'm pretty sure can be avoided.

Further down the line, if we wanted to, say, have views that used a more compact pattern of sticking the label above the value in smaller text, having the calling code all assuming a specific old-school grid model would make life harder.

Does that make sense?

cbrandtbuffalo:
> I explicitly don't want to have calling code pass in hardcoded width values because those would force more brittleness into the layout engine that I'm pretty sure can be avoided.
> 
> Further down the line, if we wanted to, say, have views that used a more compact pattern of sticking the label above the value in smaller text, having the calling code all assuming a specific old-school grid model would make life harder.
> 
> Does that make sense?

Yeah, if the new template includes the classes enabling the grid context (form-row), that makes sense. We can keep the current bootstrap code for areas that want different col sizes.

I do think tweaking the 3/9 numbers could be a desired customization in the grid model, so a hook to change that would make the new template more helpful. Maybe change to $LabelClass and $ValuesClass to avoid the specific "cols"? Or maybe a callback to set/modify those?

obra:
One thing I'm trying to do with this is to start to decouple the bootstrap-specific presentation details from the logical components. (To that end, in my local branch, I've renamed the component from "LabelValueRow" to "LabeledValue")

What got me headed in this direction was Keri mentioning that y'all might want to be able to put ticket metadata in a column next to the history.

Lifting out "show a value with a label" into its own component means that the display change necessary to go from this:

image

to this:

image

is about two lines of markup if one wants to make that change. It *should* be possible to toggle the rendering style between the different versions without changing every template showing a labeled value.

Right now, I'm at ` 76 files changed, 1048 insertions(+), 2525 deletions(-)` on my local branch and am not 100% done, but approaching being ready for a first review. 

Once this one is put to bed, the two other similar-ish cleanups I want to propose are to the wrapper around 99% of the /Elements/Submit calls in the codebase and extracting out the boilerplate for modals into a component.

obra:
@sunnavy - Making it a Component With Content was brilliant. Thank you. 

I've ported the vast majority of RT's labeled values to the new LabeledValue component. 

At this point, I'd love feedback and code review. 

Since it touches so many files, I *think* it makes sense to work to get this or some variant  possibly merged before trying to apply it to knottier bits of the codebase and to handle those as smaller PRs, but I'm open to whatever works best for you folks.

obra:
@cbrandtbuffalo @sunnavy - when you can, let me know what changes you'd like to this PR. I think the right thing to do is to be targeting whatever the next stable RT is rather than the current stable branch.

cbrandtbuffalo:
> @cbrandtbuffalo @sunnavy - when you can, let me know what changes you'd like to this PR. I think the right thing to do is to be targeting whatever the next stable RT is rather than the current stable branch.

We discussed and the downside of branching from master is that after the merge, merging up from 5.0-trunk is going to get more difficult. As long as the changes don't cause big feature shifts on 5.0, we're good with branching from 5.0-trunk. That will also allow us to get eyes on it sooner since we can start testing soon after it gets merged.

Open to other thoughts if you see any risk to going with 5.0-trunk or if you think the changes might not qualify as "stable".

obra:
i’m okay with it going on 5.0-trunk. there are some bigger cleanups that I would want to do that I don’t think makes sense for 5.0-trunk and I’m not sure how you would want to handle those.

cbrandtbuffalo:
> i’m okay with it going on 5.0-trunk. there are some bigger cleanups that I would want to do that I don’t think makes sense for 5.0-trunk and I’m not sure how you would want to handle those.

Could you do the cleanups in a separate branch from master? Or are they things you would want to do along with the other changes?

obra:
I guess my real question is where one should be targeting
not-backwards-compatible changes for 6.0?

cbrandtbuffalo:
> I guess my real question is where one should be targeting not-backwards-compatible changes for 6.0?

master is the starting point for for the next release, so you can branch from there.

obra:
Got it. once this pr is merged to 5.0-trunk and merged down to master, I’ll target bigger changes over there

cbrandtbuffalo (commented):
A previous commit, dfbcd096a4, set the ratio for the Modify Scrip page to 2/9 in an attempt to get the form entries more centered. All but Applies to gets set back to the standard 3/9, so Applies to sticks out. I'm open to a different way to make the form look a little nicer and stick with 3/9. Another approach is making the form container smaller, like on the queue configuration page.

Screen Shot 2022-08-24 at 4 06 32 PM

cbrandtbuffalo:
I made some fixes from the initial PR and rebased on top of 5.0-trunk and pushed that here:

https://github.com/bestpractical/rt/tree/5.0/extract-label-value-rows

If you could start from there with any new changes, that will pick up those fixes.

A few other areas where something is off:

* Admin > Tools > Shredder: Behavior is kind of strange, some plugin sections don't load for me when selecting different values from Select Plugin (behavior is inconsistent). I got Attachments to load and it was skewed a bit.
* The Search > Assets > Simple Search page, Basics section, has some fields not lining up
* Something on ticket display causes the columns to shift. I'm guessing it's in More about requestors as it seems to happen only if there is a requestor who isn't the logged in user
* On ticket create, on the Details tab, Links ends up at the bottom

Screen Shot 2022-08-24 at 4 16 20 PM
Screen Shot 2022-08-24 at 4 23 13 PM
Screen Shot 2022-08-24 at 4 26 49 PM
Screen Shot 2022-08-24 at 4 27 59 PM

obra:
Rebased on top of that branch. I've fixed (as far as I can tell) everything except the ticket display rendering issue you've flagged. I can't replicate that one. Can you tell me a bit more about it?

obra:
And ModifyScrip. That one is incoming.

obra:
I pushed up the change to fix 'Applies To' on scrip edit to set it back to 3/9. 

I think that custom widths for any given admin form are probably not the right optimization. We have enough similar list-of-stuff forms that we'd be better off solving it systematically.

cbrandtbuffalo:
> I pushed up the change to fix 'Applies To' on scrip edit to set it back to 3/9.
>
> I think that custom widths for any given admin form are probably not the right optimization. We have enough similar list-of-stuff forms that we'd be better off solving it systematically.

Agree.

> Rebased on top of that branch. I've fixed (as far as I can tell) everything except the ticket display rendering issue you've flagged. I can't replicate that one. Can you tell me a bit more about it?

I think it might be related to custom roles in the People section. In my test case, I had a custom role that sorts right after Owner. Applying that to a queue seems to trigger the issue, which also shows up as the People portlet rendering in edit mode.

I noticed the footer has a white gap without the custom role applied, so maybe a missing closing div or something that causes different issues if the custom role is or isn't there.

Screen Shot 2022-08-25 at 9 38 10 AM
Screen Shot 2022-08-25 at 9 39 04 AM

obra:
Looks like the issue may have been in the Links rendering. I've cleaned it
up. I think this is ready for re-review.

share/html/Ticket/Elements/EditPeopleInline
```diff
@@ -65,15 +63,9 @@
...
-  <div class="label col-3">
-    <span class="<% $role->EntryHint ? 'prev-icon-helper' : '' %>"><% $role->Name %>:</span>\
-% if ( my $hint = $role->EntryHint ) {
-<span class="far fa-question-circle icon-helper" data-toggle="tooltip" data-placement="top" data-original-title="<% $hint %>"></span>
-% }
-  </div>
-  <div class="value col-6"><& /Elements/SingleUserRoleInput, role => $role, Ticket => $Ticket &></div>
-</div>
+<&| /Elements/LabeledValue, Label => $role->Name, LabelTooltip => $role->EntryHint &>
+  <& /Elements/SingleUserRoleInput, role => $role, Ticket => $Ticket &></div>
```
cbrandtbuffalo, line 67: The stray closing div at the end of this line was causing the issues on ticket display. I'll get it in the branch I pulled in.

cbrandtbuffalo (commented):
Comment added, will take care of it when merging.

obra:
Good catch. Thank you.

cbrandtbuffalo:

PR #342, opened 2022-08-03
Minor improvements to the github testing workflows for perf, clarity and non-bestpractical use

obra:
These changes switch our github test runners to using RT's existing test-parallel target, cache timing across runs so we can run the slowest tests first, and turn off attempts at slack notifications if run from a fork outside bestpractical's github org.

The biggest possible issue to look for in this branch is the changes not triggering slack notifications when run on bestpractical/rt because I blew the conditional.


cbrandtbuffalo:
Merged, thanks!

PR #343, opened 2022-09-07
Move a UI pattern making a Submit button into a "full width" item into /Elements/Submit

obra:
This moves a bunch of style markup into a common location and should not
result in noticeable rendering changes.

### DIFF
--- a/share/html/Admin/Actions/Create.html
+++ b/share/html/Admin/Actions/Create.html
@@ -55,12 +55,9 @@
 <& Elements/EditBasics, %ARGS, Action => $action, &>
 </&>
 
-  <div class="form-row">
-    <div class="col-12">
-      <& /Elements/Submit, Label => loc('Create'), Name => 'Create' &>
-    </div>
-  </div>
 
+<& /Elements/Submit, Label => loc('Create'), Name => 'Create', FullWidth => 1 &>
+ 
 </form>
 <%ARGS>
 $Create => undef
--- a/share/html/Admin/Actions/Create.html
+++ b/share/html/Admin/Actions/Create.html
@@ -57,7 +57,7 @@
 
 
 <& /Elements/Submit, Label => loc('Create'), Name => 'Create', FullWidth => 1 &>
- 
+
 </form>
 <%ARGS>
 $Create => undef
--- a/share/html/Admin/Actions/Display.html
+++ b/share/html/Admin/Actions/Display.html
@@ -54,11 +54,7 @@
 
 <form>
 <input type="hidden" class="hidden" name="id" value="<% $id %>" />
-  <div class="form-row">
-    <div class="col-12">
-      <& /Elements/Submit, Label => loc('Copy Action'), Name => 'Copy' &>
-    </div>
-  </div>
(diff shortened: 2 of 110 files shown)


cbrandtbuffalo:
I found three places that have a special case where the buttons were right next to each other and the update adds extra divs between, creating some extra space:

share/html/Admin/Tools/GnuPG.html
share/html/Search/Build.html
share/html/Ticket/Forward.html

Easiest might be to not use the new template for these?

Screen Shot 2022-09-08 at 11 53 53 AM
Screen Shot 2022-09-08 at 11 53 41 AM

A few files report some stray whitespace (this is from git am with a patch from the branch):

Applying: Add a new 'FullWidth' option to /Element/Submit
.git/rebase-apply/patch:15: trailing whitespace.
% }
.git/rebase-apply/patch:24: trailing whitespace.
% if ($FullWidth) {
.git/rebase-apply/patch:26: trailing whitespace.
% }
warning: 3 lines add whitespace errors.
Applying: Move a UI pattern making a Submit button into a "full width" item into /Elements/Submit
.git/rebase-apply/patch:127: trailing whitespace.

.git/rebase-apply/patch:551: trailing whitespace.
    Label => loc("Update"),
.git/rebase-apply/patch:1069: trailing whitespace.

.git/rebase-apply/patch:2018: trailing whitespace.

warning: 4 lines add whitespace errors.

obra:
I'll get the trailing whitespace cleaned up.

I'd argue that the extra padding around those stacked buttons is an
improvement, as it makes it a little bit less likely to accidentally
mis-click and improves tapablity on touch devices, but if it's breaking the
current house style guide, I can revert that change in those three places.

cbrandtbuffalo:
> I'd argue that the extra padding around those stacked buttons is an improvement, as it makes it a little bit less likely to accidentally mis-click and improves tapablity on touch devices, but if it's breaking the current house style guide, I can revert that change in those three places.

We don't have an official style guide and stacked buttons don't come up much, so there aren't a ton of examples. To me it looks off because the spacing between the buttons is different from the space between the top button and the block above it. The Search button seems to be floating a bit and not completely grouped with the button above it. But Bootstrap added a bunch of spacing everywhere, so I've been sensitive to white space and trying to keep things tighter than the typical bootstrap defaults.

obra:
Digging into the CSS, it looks like that unbalanced height was caused by
this bit of css:


form div.submit div.buttons {
min-height: 2.8em;
}

That's legacy styling that predates the move to bootstrap. I'm going to
commit a proposed removal of it.

(It's from 2010 and was implemented for the back/next feature in the
installer UI which appears to work just great without it.)

[image: image.png]

commit 8cf8dcc309280a276624d82928a4c633c1258bce
Author: sunnavy ***@***.***>
Date:   Thu Dec 16 18:12:17 2010 +0800

    make "back button" after "next button" so we can submit the "next
button" when hit enter

obra:
@cbrandtbuffalo What else does this need before it's mergeable?

sunnavy:
I squashed the trailing space cleanup commit, also tweaked a few more space changes and then merged.

PR #345, opened 2022-12-15
Adding Auto-submitted: auto-generated header

dverdin:
Currently, RT can protect itself against auto-generated messages it receives.

However it does not set the "Auto-Submitted: auto-generated" header itself, though all RT messages are basically machines-generated messages.

Consequently, if a robot is set as a message receipient in RT, it will answer this messages, leading to ticket pollution.

In addition, if the robot does not seet the Auto-submitted header itself, RT will in turn answer to the robot's answer, leading to SMTP loops.

Adding the Auto-Submitted: auto-generated header gives a chance to prevent such loops.

Now, you probably have very good reasons for not setting this header already, as I am probably not the first one to report this problem. I will therefore not complain if you reject this PR. But if you have a solution to prevent the loops I was speaking about, I would gladly read it!

Thanks for this very good product.

Best regards,

### DIFF
--- a/lib/RT/Action/SendEmail.pm
+++ b/lib/RT/Action/SendEmail.pm
@@ -670,6 +670,7 @@ sub SetRTSpecialHeaders {
         }
     }
 
+    $self->SetHeader( 'Auto-Submitted', 'auto-generated' );
     $self->SetHeader( 'X-RT-Loop-Prevention', RT->Config->Get('rtname') );
     $self->SetHeader( 'X-RT-Ticket',
         RT->Config->Get('rtname') . " #" . $self->TicketObj->id() );


cbrandtbuffalo:
Hi David,

RT has one header that helps indicate the email is from a system:

https://docs.bestpractical.com/rt/5.0.3/RT_Config.html#DefaultMailPrecedence

but some auto-reply systems, like an external mail server, might not see that as you noted. We do add the header you suggest for Autoreply email, since that is truly machine generated:

https://github.com/bestpractical/rt/blob/59c0d3ca4842c5ac32d5bc11fd5844afae6bff21/lib/RT/Action/Autoreply.pm#L120

Other email from RT is created by someone replying (either in the web UI or via email), so it's human-generated even though RT sends it. So we don't include that header by default for those. Plus it can interfere with delivery since some systems will mark auto mail as SPAM.

But if you have a mail loop issue, you can always add this header to some (or all) of your templates to include it.

I'm going to close this, but feel free to re-open if you have additional thoughts. Also, if you have examples of auto-reply email from an external system that RT did not flag correctly, allowing a loop, we would definitely look at that. As you noted, we try hard to help RT recognize auto-generated email and avoid loops.

Thanks!
Jim

PR #346, opened 2022-12-16
typo

listerr:


cbrandtbuffalo:
Merged into 4.4-trunk, which will end up in the next 4.4 and 5.0 releases.

Thanks!

PR #348, opened 2023-01-10
Abstract ShowHistoryHeader title into component parameter

mzagrabe:
This allows calling components to (potentially) set the title.


sunnavy:
Merged accordingly, thanks!

-sunnavy

PR #349, opened 2023-01-12
RT_Config.pm.in - 'pririty' typo

NReilingh:
Not 100% sure which branch to open PRs against, but I inferred that 5.0-trunk might be the right one for this.

(P.S. would you consider adding a `CONTRIBUTING.md` to the repository so that there's some guidance available as to which branch(es) to file PRs against? I'd be happy to give that an initial effort.)

### DIFF
--- a/etc/RT_Config.pm.in
+++ b/etc/RT_Config.pm.in
@@ -2870,7 +2870,7 @@ queues. Values can be an ArrayRef, HashRef, or C<0>.
 
 =item ArrayRef
 
-This is the ordered String => Number map list. Pririty options will be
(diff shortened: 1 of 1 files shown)


NReilingh:
Ah, thanks for pointing me to that doc. That's the information I was looking for!

That said, you might consider a `CONTRIBUTING.md` in the GitHub default branch that does nothing more than refer the reader to `docs/hacking.pod` for convenience.  `CONTRIBUTING.md` is enough of a GitHub idiom that people like myself default to reading that file before opening their first PR instead of guessing whether they need to search for "contributing", "hacking", "pull requests", "merge requests", "patches", etc...

PR #358, opened 2023-04-13
This patch adds user "chips" which is the first step to being able to

obra:
add modern user-icons in RT's UI. The code pulls out the user's initials and renders them inline inside a rounded rectangle.

It uses one set of colors for staff and another set of colors for non-staff. This is a first pass for discussion and later improvement. I'm not in love with the colors.


obra:
image
image
image

PR #359, opened 2023-05-16
fix rt-bugs #37601 - add rt-clean-shorteners

firefart:
This fixes https://rt.bestpractical.com/Ticket/Display.html?id=37601 
The new rt-clean-shorteners was never added to the sbin path


cbrandtbuffalo:
Merged. Thanks!

PR #361, opened 2023-05-19
AutoOpenInactive bugfix and action prepare return values

NReilingh:
This PR originated as an unexpected behavior on my production RT instance. I noticed that tickets were being moved into open status unexpectedly -- for example, a new ticket that the requestor made the first reply to would be opened by the requestor's reply.

I initially thought this was a bug in `AutoOpen`, so I went looking through the code of it and other actions. I came across a sort of red herring -- that the `Prepare` guards were returning true instead of false. These actually don't have a functional effect because the `Commit` phase will return early if the last part of `Prepare` is not reached, but they do make the debug log messages more confusing since, really, `Commit` should never be run if `Prepare` fails to satisfy all of its guard clauses.

Eventually, though, I realized that `AutoOpenInactive` was running _not just_ on `inactive` statuses (but also on `initial` statuses), and so _it_ was the culprit of my "new tickets being auto-opened" problem.

I am submitting these two commits to fix the bug in `AutoOpenInactive` so that its behavior matches its documentation and name (only operating on tickets in an `inactive` status), and to clean up the `Prepare` blocks of three RT Actions so that the return value semantics are more correct.

### DIFF
--- a/lib/RT/Action/AutoOpen.pm
+++ b/lib/RT/Action/AutoOpen.pm
@@ -82,15 +82,15 @@ sub Prepare {
 
     my $ticket = $self->TicketObj;
     my $next = $ticket->FirstActiveStatus;
-    return 1 unless defined $next;
+    return 0 unless defined $next;
 
     # no change if the ticket is in initial status and the message is a mail
     # from a requestor
-    return 1 if $ticket->LifecycleObj->IsInitial($ticket->Status)
+    return 0 if $ticket->LifecycleObj->IsInitial($ticket->Status)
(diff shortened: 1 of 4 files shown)


NReilingh:
Noticed in my fork that GitHub Actions were running the test suite and failing a bunch of tests. Pretty interesting, considering that this particular action doesn't have a test suite!

Would appreciate getting some feedback on this -- I haven't yet figured out how to build and run RT from source in development mode yet (though I would like to), but I know that there is a docker harness for running the test suite at least.

PR #362, opened 2023-05-21
SetStatus docs proofreading and 'apropriate' typos

NReilingh:
I initially just noticed the "scheam" typo in SetStatus.pm and was just going to submit that, but then I noticed that the rest of the docs in the same module could use an edit pass for clarity, spelling, and grammar.

I also noticed that "apropriate" was misspelled in a few other places, so I searched it in the codebase and corrected the other user-facing instances.

### DIFF
--- a/lib/RT/Action/Autoreply.pm
+++ b/lib/RT/Action/Autoreply.pm
@@ -87,7 +87,7 @@ sub SetRecipients {
 
 =head2 SetReturnAddress
 
-Set this message's return address to the apropriate queue address
+Set this message's return address to the appropriate queue address
 
 =cut
 
--- a/lib/RT/Action/SetStatus.pm
+++ b/lib/RT/Action/SetStatus.pm
@@ -90,7 +90,7 @@ according to transitions to the target set, for example:
 
     * The current status is 'open'
     * Argument of this action is 'inactive'
-    * Possible transitions from 'open' are 'resolved', 'rejected' or 'deleted'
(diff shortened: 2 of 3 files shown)


sunnavy:
Hi Nick

I'm lean to merge the first commit and leave the 2 oxford comma commits as they are not quite necessary if you agree.

Thanks!

Regards
sunnavy

NReilingh:
@sunnavy

Nearly all formal style guides recommend the use of the serial ("Oxford") comma. Unless there's an official writing style guide for RT's docs that says something like "serial commas are discouraged", I treat them as an error when I do a proofreading/grammar pass on something.

But I don't think that's the case, since I've seen far more of the oxford comma in the rest of the writing in the docs. They should be added here to be consistent with the rest of the docs style.

sunnavy:
Hi NReilingh

I'm ok to use oxford commas and I agree consistency is important. But instead of 2 commits that fix 2 oxford commas respectively, I prefer one commit that fixes all oxford commas in the whole codebase, and then I'll be happy to merge it.

Regards
sunnavy

cbrandtbuffalo:
As a status update, I'm fine with the comma fixes for docs. We're mostly focused on 5.0 now and this doesn't apply currently. I think that's mostly because we got the spelling fixes from the other branch we received that fixed all of our spelling errors: https://github.com/bestpractical/rt/commit/464dacd1ac

So if you could branch from 5.0-trunk and include just the remaining issues/fixes we can try to merge again.

PR #364, opened 2023-06-24
CustomField.pm: mislabeled documentation header

NReilingh:
ValidateQueue is repeated twice; first instance should be ValidateHead

### DIFF
--- a/lib/RT/CustomField.pm
+++ b/lib/RT/CustomField.pm
@@ -721,7 +721,7 @@ sub ValidateValue {
     return $ret;
 }
 
-=head2 ValidateQueue Queue
+=head2 ValidateName Queue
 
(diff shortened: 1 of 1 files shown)


cbrandtbuffalo (commented):
I think Queue is also incorrect as the parameter. It should be "Name" also.

NReilingh:
@cbrandtbuffalo I meant to write "ValidateName" not "ValidateHead".

And yeah I see what you mean. The inherited method is documented as NAME.

Is there a consistent usage of CAPS for param names? Seems like the Title Casing of this set of params is anomalous. Let me add a commit and you can decide if it works for you.

cbrandtbuffalo:

PR #365, opened 2023-06-29
feat: Support for mapping remote user auth claims to user attributes

rizlas:
Hi,
this PR aim to close #208. Requested changes in #208 have been implemented. This will now support not only shibboleth but any remote user based authentication (e.g oidc) as long as claims are passed via http headers.

This settings represent the mapping between user field and http headers.
As stated in #208 it is also possible to define multiple headers that you can receive a value from (e.g. RealName).

Of course if `WebRemoteUserAuth` and `WebRemoteUserAutocreate` are disabled, `RemoteUserHeaders` will be ignored.

```
# See lib/RT/User.pm for RT possibile mapping fields
Set($RemoteUserHeaders,
    {
        'attributes_map' => {
            'Name'          => ['HTTP_USERNAME'],
            'EmailAddress'  => ['HTTP_MAIL'],
            'RealName'      => ['HTTP_DISPLAYNAME','HTTP_CN','HTTP_GIVENNAME'],
            'Gecos'         => ['HTTP_SURNAME'],
        }
    }
);
```

Apache example of setting environment variable is:

`ProxyFCGISetEnvIf "true" HTTP_DISPLAYNAME   "%{reqenv:OIDC_CLAIM_name}"`

Thank you


einar:
Hi @rizlas, nice to see someone using this. :) Just FYI I will be quite limited in helping out to get this merged as I have not touched any perl since 2017 and currently delve in infrastructure so I no longer have any access to any testing environment(s). Would be great if it could be merged though with your (or RT's) help as I still consider RT to be one of the best ticket-trackers there is and it's sad to keep this out of master/stable.

rizlas:
Hi, yeah it is really helpful for remote user based auth.
I've never written perl till today, I just tweaked your original PR 🤣 

I didn't include the utf8 fix that I read in your dockerization (https://github.com/SUNET/docker-rt-swamid/blob/master/Web.pm.patch#L47). However, in the absence of an equivalent function, I believe that that should be added too.

einar:
You do need it unless you expect only strict ascii-users (which never happens, there's always some random chars in names). :) Names will misbehave unless you keep the utf8-line.

rizlas:
> (which never happens, there's always some random chars in names)

Yeah, that's why for me it should be included 😆 But maybe there is already a helper function that does utf8 conversion.

PR #366, opened 2023-09-04
Fix pod heading

puck:
This resulted in perldoc (etc) output that looked like:

```
rt-munge-attachments
    rt-munge-attachments - Remove or replace strings in attachment records

SYNOPSIS
...
```

which I doubt is intended.

Patch in Debian: https://salsa.debian.org/request-tracker-team/request-tracker5/-/blob/master/debian/patches/fix_pod_rt_munge_attachments.diff


cbrandtbuffalo:

PR #183, opened 2016-03-18
Improve REST handling of txn CFs

sartak:
http://lists.bestpractical.com/pipermail/rt-devel/2016-March/012530.html

PR #368, opened 2023-10-20
REST 1.0 - Add fetching and setting of transaction CustomFields

puck:
These patches allow CustomFields to be set on transactions via the REST API. Thread on rt-devel is here: https://lists.bestpractical.com/pipermail/rt-devel/2016-March/012530.html

It is a resurrection of https://github.com/bestpractical/rt/pull/178 which I expect was automatically closed because the branch name started with 4.2. This patch set applies cleanly against 4.4-trunk and 5.0-trunk.

It is also the patchset from https://github.com/bestpractical/rt/pull/183 which was closed in favour of 178.

Cheers,
Andrew

### DIFF
--- a/lib/RT/Test.pm
+++ b/lib/RT/Test.pm
@@ -1000,6 +1000,33 @@ sub load_or_create_custom_field {
     return $obj;
 }
 
+=head2 load_or_create_txn_custom_field
+
+=cut
+
+sub load_or_create_txn_custom_field {
+    my $self = shift;
+    my %args = ( Disabled => 0, @_ );
+    my $obj = RT::CustomField->new( RT->SystemUser );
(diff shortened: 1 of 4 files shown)


puck:
Haha, understandable.

I was looking at what patches we were still carrying locally. We're no longer using the REST 1.0 API, so I was planning on dropping these patches, but recall that others have been using them, so thought I'd refresh the branch.

cbrandtbuffalo:
OK, cool. In the interest of stability for REST 1.0 and also keeping this code available, I'll plan to leave this branch here for anyone who might need it. If the patches apply cleanly now, they should continue to do so if we leave the REST 1.0 code alone (except for any security issues).

PR #369, opened 2023-10-28
Fix spelling in documentation

puck:
In some cases a little work on some grammar.

While building the packages for Debian, lintian complained about two common spelling mistakes in the man pages. I looked a bit further, and next thing I knew I was running ispell over all the POD. I haven't bothered trying to think about comments, although, I guess that could be pulled out and have ispell run over that as well...


cbrandtbuffalo:
Merged into 5.0-trunk. That's a lot of fixed typos, thanks!

PR #375, opened 2023-12-18
5.0/add create tickets template gui

bkembreeBPS:
There is one issue that definitely needs addressing but I wanted to discuss the best solution. When editing a template via Admin/Queues the Basics, Content, and Advanced menu options are not visible enough to the user. They need to know they are under the page menu Templates option.

The other question I had was whether we should split the Template.html into separate pages for Basics, Content, and Advanced instead of using the ContentTab/AdvancedTab url params.

### DIFF
--- a/lib/RT/Interface/Web.pm
+++ b/lib/RT/Interface/Web.pm
@@ -3171,6 +3171,115 @@ sub ProcessCustomFieldUpdates {
     return (@results);
 }
 
+=head2 ProcessTemplateUpdate ( TemplateObj => $Template, ARGSRef => \%ARGS );
+
+Returns an array of results messages.
+
+=cut
+
+sub ProcessTemplateUpdate {
+    my %args = (
+        TemplateObj => undef,
+        ARGSRef     => undef,
+        @_
+    );
+
+    my @results;
+    my @FIELDS = qw(
+        PerlCode
+        Queue
+        Subject
+        Status
+        SLA
+        Due
+        Starts
+        Started
+        Resolved
+        Owner
+        Requestor
+        Cc
+        AdminCc
+        RequestorGroup
+        CcGroup
+        AdminCcGroup
+        TimeWorked
+        TimeEstimated
+        TimeLeft
+        InitialPriority
+        FinalPriority
+        Type
+        DependsOn
+        DependedOnBy
+        RefersTo
+        ReferredToBy
+        Members
+        MemberOf
+        CustomFields
+        Content
+        ContentType
+        UpdateType
+        SkipCreate
+    );
+
+    # ensure $args{ARGSRef}{CreateSectionName} is an array ref
+    # could be undefined (new template), scalar (single section), or an array ref (multiple sections)
+    if ( $args{ARGSRef}{CreateSectionName} ) {
+        $args{ARGSRef}{CreateSectionName} = [ $args{ARGSRef}{CreateSectionName} ]
+            unless ref $args{ARGSRef}{CreateSectionName};
+    }
+    else {
+        $args{ARGSRef}{CreateSectionName} = [];
+    }
+
+    push @{ $args{ARGSRef}{CreateSectionName} }, 'ADD-NEW-SECTION'
+        if $args{ARGSRef}{AddNewSectionName};
+
+    if ( @{ $args{ARGSRef}{CreateSectionName} } ) {
+        my $new_content = '';
+        for my $name ( @{ $args{ARGSRef}{CreateSectionName} } ) {
+            my $template_name = $name eq 'ADD-NEW-SECTION' ? $args{ARGSRef}{AddNewSectionName} : $name;
+            $new_content .= "===Create-Ticket: $template_name\n";
+            for my $field ( @FIELDS ) {
+            if ( $field eq "CustomFields" ) {
+                while ( my $cf_id = shift @{ $args{ARGSRef}{ $name . "-CustomField-id" } || [] } ) {
+                my $cf_val = shift @{ $args{ARGSRef}{ $name . "-CustomField-val" } || [] };
+                    $new_content .= "CustomField-$cf_id: $cf_val\n";
+                }
+            }
+            else {
+                if ( my $val = $args{ARGSRef}{"$name-$field"} ) {
+                if ( $field eq 'PerlCode' ) {
+                    # trim leading and trailing whitespace
+                    $val =~ s/^\s+//;
+                    $val =~ s/\s+$//;
+                    $new_content .= "{\n$val\n}\n";
+                }
+                elsif ( $field eq 'Content' ) {
+                    $new_content .= "Content: $val\nENDOFCONTENT\n"
+                }
+                else {
+                    $new_content .= "$field: $val\n";
+                }
+                }
+            }
+            }
+        }
+        $args{ARGSRef}{Content} = $new_content;
+    }
+
+    my @attribs = qw( Name Description Queue Type Content );
+    my @aresults = UpdateRecordObject(
+        AttributesRef => \@attribs,
+        Object        => $args{TemplateObj},
+        ARGSRef       => $args{ARGSRef}
+    );
+    push @results, @aresults;
+
+    my ( $ok, $msg ) = $args{TemplateObj}->CompileCheck;
+    push @results, $msg if !$ok;
+
+    return ( @results );
+}
 
 
 =head2 ProcessTicketBasics ( TicketObj => $Ticket, ARGSRef => \%ARGS );
--- a/share/html/Admin/Elements/ModifyCreateTemplate
+++ b/share/html/Admin/Elements/ModifyCreateTemplate
@@ -1,0 +1,528 @@
+%# BEGIN BPS TAGGED BLOCK {{{
+%#
+%# COPYRIGHT:
+%#
+%# This software is Copyright (c) 1996-2023 Best Practical Solutions, LLC
+%#                                          <sales@bestpractical.com>
+%#
+%# (Except where explicitly superseded by other copyright notices)
+%#
+%#
+%# LICENSE:
+%#
+%# This work is made available to you under the terms of Version 2 of
+%# the GNU General Public License. A copy of that license should have
+%# been provided with this software, but in any event can be snarfed
+%# from www.gnu.org.
+%#
+%# This work is distributed in the hope that it will be useful, but
+%# WITHOUT ANY WARRANTY; without even the implied warranty of
+%# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+%# General Public License for more details.
+%#
+%# You should have received a copy of the GNU General Public License
+%# along with this program; if not, write to the Free Software
+%# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+%# 02110-1301 or visit their web page on the internet at
+%# http://www.gnu.org/licenses/old-licenses/gpl-2.0.html.
+%#
+%#
+%# CONTRIBUTION SUBMISSION POLICY:
+%#
+%# (The following paragraph is not intended to limit the rights granted
+%# to you to modify and distribute this software under the terms of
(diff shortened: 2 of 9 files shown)


share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+        my $content = $template_object->Content || '';
+        if ( substr( $content, 0, 3 ) eq '===' ) {
+
+            # parse the content and load gui controls
+            my ( $current_ticket, %current_section );
+            my @lines = split /\n/, $content;
+            while ( my $line = shift @lines ) {
+              if ( $line =~ /^===(.*)-Ticket: (.*)$/ ) {
+                # TODO: do we need to do anything different for different ticket action?
+                #       Create vs Update vs Base
```
cbrandtbuffalo, line 457: Good question. I have never used or seen examples of update-ticket or base-ticket. It would be good at some point to test those out and update docs. For the GUI, we can add an option later to select from create, update, base.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+                  $current_section{action} = $ticket_action;
+                  $current_section{active} = 1;
+                }
+                elsif ( $ticket_name ne $current_ticket ) {
+                  # new ticket template
+                  push @sections, { %current_section };
+                  %current_section         = ();
+                  $current_ticket           = $ticket_name;
+                  $current_section{name}   = $ticket_name;
+                  $current_section{action} = $ticket_action;
```
cbrandtbuffalo, line 474: Fix the spacing above.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+            # push the last template on to array
+            push @sections, { %current_section };
+        }
+        elsif ( $content ) {
+            RT->Logger->error( "'$Template' is not a Create Tickets template." );
+        }
+    }
+    else {
+        RT->Logger->error( "Couldn't load template '$Template'" );
+
```
cbrandtbuffalo, line 517: Remove extra blank line.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+      $template_object->LoadQueueTemplate( Queue => $QueueObj->id, Name => $TemplateObj->Name );
+    }
+    else {
+      $template_object->LoadGlobalTemplate( $TemplateObj->Name );
+    }
+
+    $Template = $template_object->Name;
+
+    if ( $template_object->id ) {
+        my $content = $template_object->Content || '';
```
cbrandtbuffalo, line 448: I'm wondering if we could move this parsing code to a method and make it easier to test?

share/html/Admin/Elements/ModifyTemplate
```diff
@@ -46,63 +46,81 @@
...
+        <label class="custom-control-label" for="Type-Perl"><&|/l&>Perl</&></label><br />
+      </div>
+      <div class="custom-control custom-radio">
+        <input type="radio" name="Type" class="custom-control-input" id="Type-Simple" value="Simple" <% $Type eq "Simple" ? 'checked="checked"' : "" |n %>>
+        <label class="custom-control-label" for="Type-Simple"><&|/l&>Simple</&></label><br />
+      </div>
+      <div class="custom-control custom-radio">
+        <input type="radio" name="Type" class="custom-control-input" id="Type-Create" value="Create" <% $Type eq "Create" ? 'checked="checked"' : "" |n %>>
+        <label class="custom-control-label" for="Type-Create"><&|/l&>Create</&></label><br />
+      </div>
```
cbrandtbuffalo, line 82: Change the Type to a dropdown now that we have three values.

Update the "Create" option to "Create Ticket".

share/html/Admin/Global/Template.html
```diff
@@ -56,13 +56,18 @@
 <input type="hidden" class="hidden" name="Template" value="<%$TemplateObj->Id%>" />
 % }
```
cbrandtbuffalo, line 58: In the commit message, mention "global" here and "queue" in the next commit? It makes it easier to see why there are two that look very similar.

share/html/Admin/Global/Template.html
```diff
@@ -93,6 +93,16 @@
 
 if ( $TemplateObj->Id() ) {
     push @results, ProcessTemplateUpdate( TemplateObj => $TemplateObj, ARGSRef => \%ARGS );
```
cbrandtbuffalo, line 95: There are a few areas where we're doing the same thing in global and queue level. Is it possible to factor some changes into a single mason template and avoid the duplication? If it's too much of a refactor, we don't need to do it now.

cbrandtbuffalo:
> The other question I had was whether we should split the Template.html into separate pages for Basics, Content, and Advanced instead of using the ContentTab/AdvancedTab url params.

Yes, I think separate files for each page follows the same pattern for other objects in RT.

lib/RT/Interface/Web.pm
```diff
@@ -3171,6 +3171,115 @@ sub ProcessCustomFieldUpdates {
     return (@results);
 }
 
+=head2 ProcessTemplateUpdate ( TemplateObj => $Template, ARGSRef => \%ARGS );
```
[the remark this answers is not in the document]
bkembreeBPS: At the time I thought they were alphabetical but I see now they are not. Will move it to the end.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+        </div>
+      </div>
+      </&>
+
+      <&| /Elements/LabeledValue, Label => loc('Subject'), Class => 'edit-custom-field' &>
+      <div class="form-row">
+        <div class="col-12">
+          <input name="<% $template->{name} %>-Subject" type="text" value="<% $template->{subject} || '' %>" size="50" class="form-control" />
+        </div>
+      </div>
```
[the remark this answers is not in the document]
bkembreeBPS: Do not need the size="50". I think that was left over from before I was using the title boxes to organize the page better?

I always had an intention of building an element for the input fields so I did that and added share/html/Elements/CreateTemplate/FieldInput and then made use of it on this page.

Let me reduce the file by 200 lines. Very nice!

Plus eventually we can add logic to FieldInput to make the inputs "smarter" RE handling perl code vs showing a dropdown based on the field type or similar.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+      $new_cf_count++;
+    jQuery( "#CustomFields-" + $ticket_name ).append($html);
+    jQuery('.selectpicker').selectpicker();
+  };
+</script>
+<%INIT>
+my ( $template_object, @sections, $Template );
+if ( $TemplateObj && $TemplateObj->Id ) {
+    $template_object = RT::Template->new( RT->SystemUser );
+    if ( $QueueObj ) {
```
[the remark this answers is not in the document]
bkembreeBPS: Not sure. If the user can get to this point they should have rights to the template.

Changed it to $session{'CurrentUser'}

share/html/Admin/Elements/ModifyCreateTemplate
bkembreeBPS: While looking at this I realized there was some unnecessary code here checking if the template can be loaded.

This code is getting passed a TemplateObj that was already loaded. The error checking should be at a higher level when that TemplateObj is loaded.

So this code is completely gone now.

share/html/Admin/Elements/ModifyCreateTemplate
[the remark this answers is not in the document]
bkembreeBPS: While looking at this I realized there was some unnecessary code here checking if the template can be loaded.

This code is getting passed a TemplateObj that was already loaded. The error checking should be at a higher level when that TemplateObj is loaded.

So this code is completely gone now.

bkembreeBPS:
> A few other notes from testing:
> 
>     * Change the tab type for sections to "pill" to be consistent with other places like System Configuration.

Done. 

>     * Add a titlebox section to the top where we can put some documentation. You can call the section "Create Ticket Templates". Take a shot at adding some basic docs. We can refine later.

Done.

>     * It seems like Section Name is required. We should indicate that, show an error if it is not provided, and retain any form values so the user doesn't lose their work if they forget.
>
>     * If you create a section, then click on Basics or Advanced, then click back, the Section Name field is not shown. This makes sense since it's already set and maybe we're not allowing it to be changed? But if you click on "Add New Section" tab, then back, Section Name is displayed, but it's blank. Maybe rather than hiding for existing sections, we populate it with the current value and allow it to be changed?

I made several changes for these issues:

- The Section Name for add a new section can be left blank and the ProcessTemplateUpdate will auto assign a new unique section name, indicating so in the displayed message. This saves all work and they will just need to change the section name (but could leave it as is and it will work).
- Show section name for existing sections and allow it to be changed. Add some validation and indicators that the section name for existing sections is required and do not allow the form to be submitted if empty.

 
>     * Maybe related, I was able to get in a state after creating a section that I couldn't see the Custom Fields inputs, so I couldn't add a custom field.

I could not replicate this. Thinking it is not an issue with the above changes.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+        <div class="col-12">
+          <input name="<% $template->{name} %>-SLA" type="text" value="<% $template->{sla} || '' %>" size="50" class="form-control" />
+        </div>
+      </div>
+      </&>
+
+      <&| /Elements/LabeledValue, Label => loc('Time Estimated'), Class => 'edit-custom-field' &>
+      <div class="form-row">
+        <div class="col-12">
+          <input name="<% $template->{name} %>-TimeEstimated" type="text" value="<% $template->{timeestimated} || '' %>" size="50" class="form-control" />
```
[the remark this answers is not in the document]
bkembreeBPS: Do you know offhand where we might be doing this?

If I add a custom field with a very long name it wraps the same when editing custom fields on the ticket display page.

share/html/Admin/Elements/ModifyCreateTemplate
```diff
@@ -0,0 +1,528 @@
...
+        <button class="btn btn-danger" id="<% $template->{name} %>-button" type="button" onclick="RemoveSection('<% $template->{name} %>'); return false;">Delete Section</button>
+      </div>
+%   }
+    </div>
+% }
+  </div>
+% }
+
+<script type="text/javascript">
+  function RemoveSection ($name) {
```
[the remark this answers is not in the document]
bkembreeBPS: The AddCustomField JS function has some Mason code look up the appropriate custom fields based on whether this is a global template or a queue template (now that I fixed it to do that at least).

I could refactor as much of the JS as possible into a separate file if that would be better and leave as little JS as possible in this file.

share/html/Admin/Global/Template.html
bkembreeBPS: While working on splitting Template.html into Basics.html, Content.html, and Advanced.html I got it working for page Global and Queue level templates. Could probably refactor some more to reduce duplicated code but I think it is better than it was.

bkembreeBPS: Refactored it a bit more so the Basics, Content, and Advanced pages all use the same underlying element to reduce code more.

share/html/Admin/Global/Template.html
bkembreeBPS: Due to moving both global and queue template admin to Admin/Templates these commits no longer exist.

bkembreeBPS:
> > The other question I had was whether we should split the Template.html into separate pages for Basics, Content, and Advanced instead of using the ContentTab/AdvancedTab url params.
> 
> Yes, I think separate files for each page follows the same pattern for other objects in RT.

Done.

PR #376, opened 2023-12-19
Add search filter support to assets

sunnavy:


share/html/Elements/SearchFilter
```diff
@@ -56,7 +56,8 @@
           </a>
         </div>
         <div class="modal-body">
-%       if ( $Attribute eq 'Subject' ) {
+%       if ( $Class eq 'RT::Tickets' ) {
+%         if ( $Attribute eq 'Subject' ) {
```
cbrandtbuffalo, line 60: Looking at this file, I'm thinking it probably should have been originally created in share/html/Search/Elements. Do you recall why we put it in share/html/Elements?

To make this more readable, and thinking about eventually adding transactions, I wonder if we should put all of the RT::Tickets code in a new template share/html/Search/Elements/SearchFilterTickets, and RT::Assets in share/html/Search/Elements/SearchFilterAssets. Thoughts?

cbrandtbuffalo (changes requested):

share/html/Elements/SearchFilter
sunnavy: I think the initial reason we put it to "/Elements/SearchFilter" was that it's only called by "/Elements/CollectionAsTable/Header", but I agree that "/Search/Elements/" is more appropriate and I'll move it accordingly.

+1 to use different SearchFilter components

PR #377, opened 2023-12-19
5.0/asset inline edit

sunnavy:


cbrandtbuffalo (changes requested):
Some questions on a few portlets:

* Links gets squashed in the 3-column view, making the labels wrap poorly. We should make it single column with existing values on top and edit on the bottom like tickets.
* I'm seeing a pencil for dates, but when I click there's nothing to edit because the core dates for assets (Created and LastUpdated) can't be edited. Maybe we should only show the pencil if there are CFs that can be edited? Otherwise it's confusing to be able to click the pencil but not be able to edit anything.
* I'm not seeing a pencil for custom fields. Something in my config?

sunnavy:
> Some questions on a few portlets:
> 
> * Links gets squashed in the 3-column view, making the labels wrap poorly. We should make it single column with existing values on top and edit on the bottom like tickets.
> * I'm not seeing a pencil for custom fields. Something in my config?

Good catches, both are fixed.

> * I'm seeing a pencil for dates, but when I click there's nothing to edit because the core dates for assets (Created and LastUpdated) can't be edited. Maybe we should only show the pencil if there are CFs that can be edited? Otherwise it's confusing to be able to click the pencil but not be able to edit anything.

There is a callback in "/Asset/Elements/EditDates" where users can do anything, so checking cfs might not be enough.
I chose to disable inline edit by default for "Dates" widget, and users can always update config to explicitly enable it.

PR #379, opened 2023-12-20
5.0/asset inline edit

sunnavy:


cbrandtbuffalo:
I found one other strange thing that might be unrelated to the new inline code. Date CFs that have a value show "(no value)" when editing them. I see it on the Basics tab also, which makes me think it was there already. Either way, we could fix it in the branch. See attached images.
Screen Shot 2024-01-08 at 3 29 43 PM
Screen Shot 2024-01-08 at 3 29 34 PM
Screen Shot 2024-01-08 at 3 29 17 PM

cbrandtbuffalo:
Turns out it was a misconfiguration of that custom field. I had changed the types a few times and it was a multi-value, which then didn't populate the default value. When I changed it to a single date CF it was working fine.

Merged into 5.0-trunk. Assets now have inline edit, yay!

PR #384, opened 2024-01-31
fixing localisation

PPetky:
This is a fix of interface missing localisation for Search/Bulk.html and 
adding localisation to lib/RT/Interface/Web/ReportsRegistry.pm

### DIFF
--- a/lib/RT/Interface/Web/ReportsRegistry.pm
+++ b/lib/RT/Interface/Web/ReportsRegistry.pm
@@ -98,10 +98,7 @@ Every element is a hash ref with the following keys:
 
 =cut
 
-sub loc { HTML::Mason::Commands::loc( @_ ); }
-
 sub Reports {
-    while( my($k, $v) = each %$registry ) { $v->{'title'} = loc( $v->{'title'} ); }
     my @res
         = sort { lc( $a->{title} ) cmp lc( $b->{title} ) } values %$registry;
     return \@res;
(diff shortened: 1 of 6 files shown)


lib/RT/Interface/Web/ReportsRegistry.pm
```diff
@@ -98,7 +98,10 @@ Every element is a hash ref with the following keys:
 
 =cut
 
+sub loc { HTML::Mason::Commands::loc( @_ ); }
+
 sub Reports {
+    while( my($k, $v) = each %$registry ) { $v->{'title'} = loc( $v->{'title'} ); }
```
sunnavy, line 104: localising title here is kinda too early and will cause RT to save localised data to db, which is bad as users may see sticky translations even if they change the language.

Could you do the localisation later instead, e.g. in lib/RT/Interface/Web/MenuBuilder.pm and pages share/html/Prefs/DashboardsInMenu.html, share/html/Admin/Global/DashboardsInMenu.html, and share/html/Prefs/DashboardsInMenu.html?

For the latter 3 pages, I think localising the "label" item will be enough.

Other changes look good.

Thanks!

sunnavy (changes requested):

lib/RT/Interface/Web/ReportsRegistry.pm
PPetky: hi, I have made fixes... thnx

sunnavy:
I tweaked and squashed the commits and merged, thanks!

One noticeable change I made is to drop the localisation of dashboard names, which are not supposed to be localised as they are generally user-defined.

-sunnavy

PR #387, opened 2024-05-23
Use always "CustomField" attribute in header orderby links

elacour:
"CustomFieldView" is not a valid attribute for OrderBy.


sunnavy:
I fixed this issue in share/html/Elements/ColumnMap instead in 7ce5fd60e0.

Thanks!
sunnavy

PR #389, opened 2024-05-23
Always quote search value when using "LIKE" op on numerical CF

elacour:
Without this, on numerical customfield, SQL produced is wrong and crashes (at least on MySQL):

[...]ObjectCustomFieldValues_1.Content+'0') LIKE %'2'% [...]


sunnavy:
Squashed commits and merged. It will be included in 5.0.7.

Thanks!
sunnavy

PR #391, opened 2024-09-02
Fix LOCAL_PLUGIN_PATH

MikaelUrankar:
I don't know why LOCAL_PLUGIN_PATH is set to RT_LOCAL_PATH_R but it generates weird path on FreeBSD


MikaelUrankar:
Yes:
```
ls /wrkdirs/usr/ports/www/rt50/work/stage/usr/local/share/rt50/local/
etc     html    lib     po
```

cbrandtbuffalo:
I don't know all of the details of how FreeBSD ports work. It looks like this patch file might be applied as part of the build process and it is setting the directory?
https://cgit.freebsd.org/ports/tree/www/rt50/files/patch-Makefile.in

It looks like this patch is also updating the directories:
https://cgit.freebsd.org/ports/tree/www/rt50/files/patch-config.layout

But the paths are all updated to use datadir and they all remove "local" from the path. So I wonder how it gets set for the directories that are working? What does this section of the generated Makefile look like (the below is from an inplace config)?

RT_PATH			=	.
RT_ETC_PATH		=	./etc
RT_BIN_PATH		=	./bin
RT_SBIN_PATH		=	./sbin
RT_LIB_PATH		=	./lib
RT_MAN_PATH		=	./man
RT_VAR_PATH		=	./var
RT_DOC_PATH		=	./docs
RT_FONT_PATH		=	./share/fonts
RT_LEXICON_PATH		=	./share/po
RT_STATIC_PATH		=	./share/static
RT_LOCAL_PATH		=	./local
LOCAL_PLUGIN_PATH	=	./local/plugins
LOCAL_ETC_PATH		=	./local/etc
LOCAL_LIB_PATH		=	./local/lib
LOCAL_LEXICON_PATH	=	./local/po
LOCAL_STATIC_PATH	=	./local/static
MASON_HTML_PATH		=	./share/html
MASON_LOCAL_HTML_PATH	=	./local/html
MASON_DATA_PATH		=	./var/mason_data
MASON_SESSION_PATH	=	./var/session_data
RT_LOG_PATH		=       ./var/log

cbrandtbuffalo:
> Or maybe I'm expecting that LOCAL_PLUGIN_PATH should be the value from "customplugindir" but it's 2 different things?

The value for "customplugindir" should end up in LOCAL_PLUGIN_PATH, so I think that is correct.

Looking more closely at the build process and your PR, I now see the RT_LOCAL_PATH_R -> LOCAL_PLUGIN_PATH_R change. Previously I just saw the removal of /plugins. Now I'm thinking your change is correct. I'll get some additional review and comment back.

cbrandtbuffalo:
Looking at this commit: https://github.com/bestpractical/rt/commit/ac8ab44111cb6a9695642eaa339390823f051c02
I think the Makefile.in change was just missed, and FreeBSD is the only section in config.layout that has a different value, so it wasn't noticed.

So this PR is correct. I'll work on getting it merged with a little more detail in the commit message.

cbrandtbuffalo:
Merged into 5.0-trunk, so it will be in 5.0.8. Thanks!

PR #392, opened 2024-09-27
Change free port detection to how PSGI binds to a port

puck:
The previous method using socket/connect would allow us to bind to a port that PSGI then couldn't bind to. If a port is connected on a specific IP, then using connect with 0.0.0.0 would still connect okay.

Using IO::Socket::INET this will fail, which is reasonable for 0.0.0.0, and then PSGI wouldn't be able to start and the test would fail.

This may resolve the intermittant test failures.

Report in RT at: https://rt.bestpractical.com/Ticket/Display.html?id=37886
Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1074781

### DIFF
--- a/lib/RT/Test.pm
+++ b/lib/RT/Test.pm
@@ -69,6 +69,7 @@ my $Test_NoWarnings_Catcher = $SIG{__WARN__};
 my $check_warnings_in_end   = 1;
 
 use Socket;
+use IO::Socket::INET;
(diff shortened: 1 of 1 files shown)


cbrandtbuffalo:
The change makes sense. In testing, if the port isn't available it hangs for a bit, but unfortunately the Timeout argument doesn't seem to help.

PR #393, opened 2024-11-12
fix: importer didn't quote the Groups table for MySQL8

mehgcap:
Importer.pm and RightsInspector.pm both used unquoted Groups as a table name. I added the backticks (\`Groups\`) so MySQL8 won't choke on this table name. More work may need to be done to fully support other database types.


sunnavy:
Merged accordingly, thanks!

-sunnavy

PR #395, opened 2024-12-14
Add name attribute to Create New Ticket button

mzagrabe:
Add name attribute so Callbacks can divine whether the button was clicked upon form submission.


sunnavy:
Merged updated 396, closing.

-sunnavy

PR #398, opened 2025-02-10
Queues only get CustomRoles that apply to tickets

btb:
/Admin/Queues/CustomRoles.html currently always shows all custom roles regardless of whether they apply to tickets or not, even showing them as applied, if they are applied to another object type (i.e. asset catalog) which has the same id.

I followed the lead of other code dealing with CustomRoles, and (ab?)used the CustomFieldLookupType method to get the applicable LookupType.

This hard-codes EditCustomRoles to only work for administering queues, but that's all it's currently used for anyway.

### DIFF
--- a/share/html/Admin/Elements/EditCustomRoles
+++ b/share/html/Admin/Elements/EditCustomRoles
@@ -145,10 +145,12 @@ if ( $UpdateCRs ) {
 $m->callback( CallbackName => 'UpdateExtraFields', Results => \@results, Object => $Object, %ARGS );
 
 my $added_crs = RT::CustomRoles->new( $session{'CurrentUser'} );
+$added_crs->LimitToLookupType(RT::Ticket->CustomFieldLookupType);
 $added_crs->LimitToObjectId($id);
 $added_crs->ApplySortOrder;
 
 my $not_added_crs = RT::CustomRoles->new( $session{'CurrentUser'} );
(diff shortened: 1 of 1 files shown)


cbrandtbuffalo:
Hi @btb , thanks for the PR! Since custom roles can be applied to assets, I extended your PR to cover catalogs also. Could you check and see if this branch works for you? https://github.com/bestpractical/rt/tree/5.0/queue-custom-roles

btb:
That works great as well. I do really appreciate having more consistency between tickets and asses.

Speaking of which, back when I used and worked on the AssetTracker plugin, we had scrips for assets, and I liked it a great deal - I just started figuring out what would be needed to do that with RT's Assets, without adding more tables like AT did. So far it's not really been that difficult - give RT::Scrip the "RT::LookupType" role (one new column for LookupType), and let RT::Templates belong to other objects besides Queues (one new column for ObjectType), and then it kinda writes itself. I'm thinking of starting a pull request for this, I'd be happy to get feedback, if you think it would be welcomed. I wouldn't have any expectation of it being a quick turnaround of course.

btb:
> Yes, we definitely want to add scrips to assets, we just haven't gotten there yet. If you want to work on a PR, that would be great.

My work is now up at #399

cbrandtbuffalo:
This was tweaked a bit more and is now merged into 5.0-trunk. Thanks!

PR #399, opened 2025-02-26
Enable scrips for assets and articles

btb:
This is a work-in-progress for adding scrips to assets and other object types.

This is my first significant RT contribution in a while, but I've been maintaining and customizing RT as part of my job on-and-off for about 20 years (?!). I was a co-maintainer of the old AssetTracker plugin and wrote a few other ones that we use internally.

The central idea of this PR is that RT::Scrip takes on the LookupType role, and it can then operate on different object types based on that. That basic functionality is fully implemented here, and seems to be working well.

I would love if you were able to give feedback on what I've got so far, and also especially your thoughts on these remaining issues (I'm sure there will be others):

- For now I've left the variable and argument names alone ("$queue", "QueueObj", "TicketObj", etc) to keep the diff smaller. I guess they'll mostly just be changed to "Object", "ObjectId", etc unless there's a better name that's more specific, yet type-independent (ContainerObj, RecordObj, ?). There are also still some logging and localized messages that refer to "queue #" etc.

- User-defined code works fine, but most of the modules for Conditions and Actions are not going to be friendly for objects other than tickets. Options to deal with that:

  - Most Actions and Conditions could be modified, I think, so they'll work with any object, or to bail out gracefully if they can't.

  - RT::ScripAction and RT::ScripCondition each add an ObjectType column, or maybe the whole LookupType role, then each entry in those tables will tell you if it applies to Tickets, Assets, etc. If a particular ExecModule is written right, it can be applied to multiple types (just another entry in the table). When configuring a scrip, only those with the right type are available.

- Currently scrips for other objects are forced to global templates only. Ideas of how to add overrides:

  - Just add an ObjectType column, and rename Queue to ObjectId. I've noticed there's some pretty complex querying going on around what templates are used by what, these will probably get even more complex.

  - Have a whole new table for ObjectTemplates. Probably overkill in the current paradigm where non-global templates just "belong" to their container object (queue), but I have definitely thought it would be nice sometimes to apply the same template to multiple queues. Might make those complex queries simpler.

- Needs UI for editing/ordering global scrips - not sure if it'd be better to make a new page for each object type similar to /Admin/Global/CustomFields/... or to just make the existing page /Admin/Global/Scrips.html support all different types.


cbrandtbuffalo:
> This is a work-in-progress for adding scrips to assets and other object types.
> 
> This is my first significant RT contribution in a while, but I've been maintaining and customizing RT as part of my job on-and-off for about 20 years (?!). I was a co-maintainer of the old AssetTracker plugin and wrote a few other ones that we use internally.
> 
> The central idea of this PR is that RT::Scrip takes on the LookupType role, and it can then operate on different object types based on that. That basic functionality is fully implemented here, and seems to be working well.
> 
> I would love if you were able to give feedback on what I've got so far, and also especially your thoughts on these remaining issues (I'm sure there will be others):

This is a really good start. Good clean commits with good commit messages, which is great!

Sorry for the delay in reviewing. We've been hard at work on RT 6, which is coming out soon. Given that, my first thought is a big new change like this fits better in newer RT, which will be RT 6. Could you take a shot at moving this over to the master branch? I think most or all of the commits will apply cleanly given the code you are working in. The UI parts at the end might need some extra work.

>     * For now I've left the variable and argument names alone ("$queue", "QueueObj", "TicketObj", etc) to keep the diff smaller. I guess they'll mostly just be changed to "Object", "ObjectId", etc unless there's a better name that's more specific, yet type-independent (ContainerObj, RecordObj, ?). There are also still some logging and localized messages that refer to "queue #" etc.

Understood. This is the time to make changes like that, so we might drop one more commit in to do these conversions. Otherwise they end up never getting done.

>     * User-defined code works fine, but most of the modules for Conditions and Actions are not going to be friendly for objects other than tickets. Options to deal with that:
>       
>       * Most Actions and Conditions could be modified, I think, so they'll work with any object, or to bail out gracefully if they can't.
>       * RT::ScripAction and RT::ScripCondition each add an ObjectType column, or maybe the whole LookupType role, then each entry in those tables will tell you if it applies to Tickets, Assets, etc. If a particular ExecModule is written right, it can be applied to multiple types (just another entry in the table). When configuring a scrip, only those with the right type are available.

The above sounds right to me. I can definitely see many Conditions and Actions working for multiple objects, like notifications, status changes, custom field changes, etc.

>     * Currently scrips for other objects are forced to global templates only. Ideas of how to add overrides:
>       
>       * Just add an ObjectType column, and rename Queue to ObjectId. I've noticed there's some pretty complex querying going on around what templates are used by what, these will probably get even more complex.
>       * Have a whole new table for ObjectTemplates. Probably overkill in the current paradigm where non-global templates just "belong" to their container object (queue), but I have definitely thought it would be nice sometimes to apply the same template to multiple queues. Might make those complex queries simpler.

I agree. I think for this branch it's a good idea to start with global templates. Then a future branch can tackle templates. There are other changes we want to make with templates as well.

>     * Needs UI for editing/ordering global scrips - not sure if it'd be better to make a new page for each object type similar to /Admin/Global/CustomFields/... or to just make the existing page /Admin/Global/Scrips.html support all different types.

I would follow the pattern from queues and have a scrips page per object that allows ordering just for scrips applied to that object.

Thanks for all of your work so far!

btb:
rebased against master with minimal changes needed. my original version is now at [5.0/asset-article-scrips](https://github.com/btb/rt/tree/5.0/asset-article-scrips)

cbrandtbuffalo:
@btb A version of this has been merged into 6.0-trunk. Let us know if you test and find any issues. We still have time to update before RT 6 is released. Thanks for all of your work!

PR #400, opened 2025-04-15
6.0/i18n fix

elacour:
Add/fix missing i18n for RT 6.0

### DIFF
--- a/share/po/ar.po
+++ b/share/po/ar.po
@@ -18,7 +18,7 @@ msgstr ""
 "X-Generator: Launchpad (build e76edd883483c71c468bb038e98836435de44530)\n"
 
 #. ($auth_string)
-#: lib/RT/Interface/Web.pm:5693
+#: lib/RT/Interface/Web.pm:5699
 msgid "\"%1\" is your new authentication token. Treat it carefully like a password. Please save it now because you cannot access it again."
 msgstr ""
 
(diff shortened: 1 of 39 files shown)


sunnavy:
I tweaked commits a bit, the most noticeable change is I removed changes of "LabelForRole". The reason is: just like custom fields, currently we translate core roles only. I will create another branch to implement it.

In the future, we might support to translate custom fields and custom roles, once we have a good way for users to easily add translations.

elacour:
> I tweaked commits a bit, the most noticeable change is I removed changes of "LabelForRole". The reason is: just like custom fields, currently we translate core roles only. I will create another branch to implement it.

but this is consistent with the way it works on tickets. On RT 6 (and 5 too), create a CustomRole named "Save" and go to Search Builder or add watcher on a ticket, in french you will see "Sauver", not "Save". I understand it's maybe not desirable but it's the way it works as of now. On assets, up to now, we do not translate standard roles (HeldBy, ..) at all, which is very bad for foreign RT users. Hence I would suggest accepting this change until we find a proper solution for both assets and tickets customroles :)

sunnavy:
Actually, custom roles don't behave consistently on tickets right now (sorry about it), e.g. on Ticket display page, custom roles are not translated.

To translate asset core roles, I made a branch 6.0/loc-asset-core-roles, and hopefully it will be included in 6.0.0.

Thanks!

-sunnavy

PR #401, opened 2025-06-23
Add missing WebPath for modify scheduled process

lmzke:
Backport a fix for scheduled processes.


cbrandtbuffalo:
Merged into 5.0-trunk. Thanks for the fix!

PR #405, opened 2025-08-27
Make dev easier

puck:
Provide guidance on how to run a standalone dev environment.

Make rt-server having working CSRF when running in a dev environment.

### DIFF
--- a/docs/hacking.pod
+++ b/docs/hacking.pod
@@ -166,6 +166,58 @@ line) is to be loc'd.  This is useful for inline hashes:
 
 =head2 Setting up a development environment
 
+A quick and easy way to try out changes in RT is to use SQLite and
+the inplace layout so we use the code inside the git checkout. Here
+is a sample configuration you can use, put this in a file under
+C<etc/RT_SiteConfig.d>, say C<etc/RT_SiteConfig.d/dev.pm>:
+
+    Set( $rtname, 'example.com');
+    Set( $CommentAddress, 'rt-comment@example.com' );
+    Set( $CorrespondAddress, 'rt@example.com' );
+    Set( $Organization, 'Example Co' );
+
+    # Set a domain you can hit locally, for example localhost.
+    Set( $WebDomain, 'localhost' );
+    Set( $WebPort, '8080' );
+
+    # Allows you to run over http without SSL.
+    Set($WebSecureCookies, 0);
+
+    # Let's you see all RT log output in your terminal.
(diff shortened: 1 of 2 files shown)


cbrandtbuffalo (changes requested):
What do you think about adding some more docs here with a helpful dev config file? Maybe that would also cover the WebBaseURL issue. Something like:

There are some configuration settings that make development easier. The options below are a good base. If you add them to etc/RT_SiteConfig.pm, git will want to include that file in your changes. If you put these in etc/RT_SiteConfig.d/RT_SiteConfig.pm, git will ignore them.

```
Set( $rtname, 'example.com');
Set( $CommentAddress, 'rt-comment@example.com' );
Set( $CorrespondAddress, 'rt@example.com' );
Set( $Organization, 'mycomputer.local' );
# Set a domain you can hit locally, could be localhost
Set( $WebDomain, 'mycomputer.local' );
Set( $WebPort, '8080' );
# Allows you to run over http without SSL
Set($WebSecureCookies, 0);
# Let's you see all RT log output in your terminal
Set( $LogToSTDERR, 'debug');
# Send email output to files in var so you can look at them
Set( $MailCommand, 'mbox');
# Don't use the Mason cache, so you can see changes immediately
Set( $DevelMode, 1);
# Useful if you need to see raw SQL, but don't leave on because it logs a ton
#Set( $StatementLog, 1);
```

puck:
Cool, thank you. I've updated HACKING.pod to use the example.

I'm okay with dropping the patch to lib/RT/PlackRunner.pm . I guess that having it run on localhost isn't the only case, but it is nice that you can run RT without any other configuration with this patch.

cbrandtbuffalo:
> Cool, thank you. I've updated HACKING.pod to use the example.
> 
> I'm okay with dropping the patch to lib/RT/PlackRunner.pm . I guess that having it run on localhost isn't the only case, but it is nice that you can run RT without any other configuration with this patch.

OK! Merged into 5.0-trunk with a few tweaks. Thanks for the PR!

puck:
I'm just writing up some instructions for someone to test something, and realised it doesn't mention the username/password to use. It'd be good to add this as well.

cbrandtbuffalo:
> I'm just writing up some instructions for someone to test something, and realised it doesn't mention the username/password to use. It'd be good to add this as well.

It's in the README, and it could be in the dev docs also. Although it's a good idea for people to check the README also when getting set up.

PR #408, opened 2025-09-30
Fix handling of 'Auto-Submitted' mail header when set to no 

danpoltawski:
When the `Auto-Submitted` mail header is set, it is always interpreted as the mail being auto-generated. This was causing us issues with another ticketing system mailing into RT and setting the header as `Auto-Submitted: no` and RT treating it as an autogenerated

The code is clearly trying to respect this `no` value but the implementation is broken. I've added some unit test cases to demonstrate the issue.

Note that I'm not much of a Perl guru so there may be a more perl-y way to do the fix.


danpoltawski:
Not sure what I need to do to try and get bugs fixed in RT 😭 

https://rt.bestpractical.com/Ticket/Display.html?id=38069

cbrandtbuffalo:
> Not sure what I need to do to try and get bugs fixed in RT 😭
> 
> https://rt.bestpractical.com/Ticket/Display.html?id=38069

Hi Dan, you're doing the right thing and we do see your PR. We're just very busy right now with other things and haven't had time to review it. I'm assuming you have patched locally, so at least you have your immediate problem solved in your system. Thanks for the PR, and we'll get to it when we get some time!

cbrandtbuffalo:
> When the `Auto-Submitted` mail header is set, it is always interpreted as the mail being auto-generated. This was causing us issues with another ticketing system mailing into RT and setting the header as `Auto-Submitted: no` and RT treating it as an autogenerated
> 
> The code is clearly trying to respect this `no` value but the implementation is broken. I've added some unit test cases to demonstrate the issue.
> 
> Note that I'm not much of a Perl guru so there may be a more perl-y way to do the fix.

Looking at the existing code and the fix, it seems like the bug is that the "no" value returned from MIME::Head has a newline. The docs have a note that returned header values may have a newline, in the "get" section: https://metacpan.org/pod/MIME::Head#Getting%2Fsetting-fields . So I think just the "chomp" fixes the issue. I tried that and it passes the new tests. Are you able to try that with an email from the other ticketing system where you saw the issue?

PR #409, opened 2025-10-27
Support WWW::Mechanize v2.20

puck:
The select method now requires the second element to be the occurrence of the form to select. As it now supports working on not just the first form that matches. See:

  https://metacpan.org/dist/WWW-Mechanize/changes

Fixes: https://rt.bestpractical.com/Ticket/Display.html?id=38076


cbrandtbuffalo:
Merged into 5.0-trunk, which will eventually get merged up to 6.0-trunk.
Thanks for tracking this down and submitting the PR!

PR #413, opened 2026-02-13
Use always internal parsing for iso date/times

elacour:
When using RT "unknown" date format, external perl modules are used and fails:
    * Time::ParseDate fail to parse iso format and produce wrong date/time,
    * DateTime::Format::Natural claims string isn't parseable.

This is a problem when updating RT DateTime CF from REST where iso8601 is recommended. _CanonicalizeValueDateTime parse input as "unknown" and produces wrong dates.

### DIFF
--- a/lib/RT/Date.pm
+++ b/lib/RT/Date.pm
@@ -199,7 +199,7 @@ sub Set {
         $RT::Logger->warning("Invalid date $args{'Value'}: $@") if $@ && !$u;
         return $self->Unix( $u > 0 ? $u : 0 );
     }
-    elsif ( $format eq 'iso' &&
+    elsif ( ( $format eq 'iso' || $format eq 'unknown' ) &&
           ( $args{'Value'} =~ /^(\d{4})-(\d\d)-(\d\d)[ T](\d\d):(\d\d):(\d\d)Z?$/ ||
             $args{'Value'} =~ /^(\d{4})(\d\d)(\d\d)T(\d\d)(\d\d)(\d\d)Z?$/)) {
         local $@;


elacour:
Before this PR:


$ ./test.pl 
Testing with "2026-02-13T17:16:35Z"

Parsed as "unknown" with Time::ParseDate: 2026-02-12 23:00:00

[900183] [Fri Feb 13 17:38:23 2026] [warning]: Couldn't parse date '2026-02-13T17:16:35Z' by DateTime::Format::Natural
Parsed as "unknown" with DateTime::Format::Natural: 2026-02-12 23:00:00

Parsed as "iso": 2026-02-13 17:16:35

elacour:
Maybe adding a chapter on REST2 doc about doc formats would be nice ?

cbrandtbuffalo:
> * Time::ParseDate fail to parse iso format and produce wrong date/time,

I'm inclined to try to get a fix in Time::ParseDate rather than covering up a bug in the RT code. There is actually a PR with a fix: https://github.com/muir/Time-modules/pull/10, but the module is looking for a new maintainer. I'll look into that and see what the options are.

cbrandtbuffalo:
@elacour We adopted Time::ParseData and published a new version with a fix for the iso format issue. Could you update that module and let us know if it fixes the issue for you?

elacour:
> @elacour We adopted Time::ParseData and published a new version with a fix for the iso format issue. Could you update that module and let us know if it fixes the issue for you?

Doesn't seems to be ok, it parses as localtime not as UTC when there is an ending Z:

Test with Europe/Paris (2026-02 = GMT+1)

Using date: 2026-02-13T17:16:35Z
With Time::ParseDate:
Ven 13 Fév 2026 17:16:35
With DateTimeFormatNatural:
Ven 13 Fév 2026 18:16:35

cbrandtbuffalo:
Yes, you're right that it still needs another change because the datetimes with the Z should not adjust for timezone. That fix has been merged in RT 6 here and will be out in the next release:
https://github.com/bestpractical/rt/commit/551d83e1b0bff6b3781b0f53e4cd86a6be2a0a3d

